PR #35, opened 2022-03-17
Suppress the difference in the state file for  labels

### LINKED ISSUE #30: Incorrectly identifying cluster state change
### Describe the bug

I created a cluster where I did not specify the following within my node pool.

            cloud_label = {}
            node_label  = {}

The cluster created fine, however each time the cluster state is refreshed for successive "applies" terraform keeps saying that there is a difference been my state and the remote implementation because of this.

I don't think this should happen.

### Reproduction steps

```bash
1. Create cluster without the labels and apply
2. Run apply again, and you will see the diff
```

### Expected behavior

I did not change the definition, so I did not expect the state change.

### Additional context

_No response_

shobha2626:
1. **What this PR does / why we need it**:  Suppress the difference in the state file for  labels

2. **Which issue(s) this PR fixes**. 

        Fixes #30 

3. **Additional information**

4. **Special notes for your reviewer** :  Suppressing the difference irrespective of the change

### DIFF
--- a/internal/resources/cluster/tkgservicevsphere/resource_tkg_service_vsphere.go
+++ b/internal/resources/cluster/tkgservicevsphere/resource_tkg_service_vsphere.go
@@ -394,12 +394,18 @@ var nodePoolSpec = &schema.Schema{
 				Description: "Node labels",
 				Optional:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return true
+				},
 			},
 			cloudLabelKey: {
 				Type:        schema.TypeMap,
 				Description: "Cloud labels",
 				Optional:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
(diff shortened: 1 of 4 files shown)


internal/resources/cluster/tkgservicevsphere/resource_tkg_service_vsphere.go
```diff
@@ -394,12 +394,18 @@ var nodePoolSpec = &schema.Schema{
 				Description: "Node labels",
 				Optional:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
shreyassreenivas, line 397: what about the new changes coming in ? will that be suppressed?

shreyassreenivas: We should not be suppressing the new changes coming in from the Tf config file

shobha2626:
There are no changes in the labels, including cloud and node labels found after the processing state of the cluster.

image

PR #39, opened 2022-04-03
Remove cloud and node labels in TKGm vSphere clusters

### LINKED ISSUE #30: Incorrectly identifying cluster state change

shobha2626:
1. **What this PR does / why we need it**: Remove cloud and node labels in TKGm vSphere

        Fixes #30

3. **Additional information**

Node and Cloud labels and not supported for TKGm vSphere workload clusters and some minor test fixes

### DIFF
--- a/internal/client/errors/errors.go
+++ b/internal/client/errors/errors.go
@@ -16,12 +16,6 @@ type ClientErrors struct {
 	err      error
 }
 
-func Error(err error) ClientErrors {
-	return ClientErrors{
-		err: err,
-	}
-}
-
 func ErrorWithHTTPCode(httpCode int, err error) ClientErrors {
 	return ClientErrors{
 		httpCode: httpCode,
--- a/internal/resources/cluster/helper_test.go
+++ b/internal/resources/cluster/helper_test.go
@@ -60,6 +60,7 @@ func withTKGsCluster() testAcceptanceOption {
 func withTKGmVsphereCluster() testAcceptanceOption {
 	return func(config *testAcceptanceConfig) {
 		config.ManagementClusterName = os.Getenv("MANAGEMENT_CLUSTER")
+		config.ProvisionerName = os.Getenv("PROVISIONER_NAME")
 		config.ControlPlaneEndPoint = os.Getenv("CONTROL_PLANE_ENDPOINT")
 		config.accTestType = tkgVsphereCluster
 		config.templateData = testTKGmVsphereClusterScript
--- a/internal/resources/cluster/tkgvsphere/resource_tkgvsphere.go
+++ b/internal/resources/cluster/tkgvsphere/resource_tkgvsphere.go
@@ -11,7 +11,6 @@ import (
 
 	nodepoolmodel "github.com/vmware-tanzu/terraform-provider-tanzu-mission-control/internal/models/cluster/nodepool"
 	tkgvspheremodel "github.com/vmware-tanzu/terraform-provider-tanzu-mission-control/internal/models/cluster/tkgvsphere"
-	"github.com/vmware-tanzu/terraform-provider-tanzu-mission-control/internal/resources/common"
 )
 
 var TkgVsphereClusterSpec = &schema.Schema{
@@ -571,18 +570,6 @@ var tkgVsphereNodePoolSpec = &schema.Schema{
 				Description: "Count is the number of nodes",
 				Optional:    true,
 			},
-			nodeLabelKey: {
-				Type:        schema.TypeMap,
-				Description: "Node labels",
-				Optional:    true,
-				Elem:        &schema.Schema{Type: schema.TypeString},
-			},
-			cloudLabelKey: {
-				Type:        schema.TypeMap,
-				Description: "Cloud labels",
-				Optional:    true,
-				Elem:        &schema.Schema{Type: schema.TypeString},
-			},
 			tkgVsphereKey: {
 				Type:        schema.TypeList,
 				Description: "Nodepool config for tkgm vsphere",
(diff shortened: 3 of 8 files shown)


Vasundharashukla (commented):
Please make changes to the concerned flattening tests as well.

internal/resources/cluster/helper_test.go
```diff
@@ -60,6 +60,7 @@ func withTKGsCluster() testAcceptanceOption {
 func withTKGmVsphereCluster() testAcceptanceOption {
 	return func(config *testAcceptanceConfig) {
 		config.ManagementClusterName = os.Getenv("MANAGEMENT_CLUSTER")
+		config.ProvisionerName = os.Getenv("PROVISIONER_NAME")
```
Vasundharashukla, line 63: Why have we added this as an environment variable? We can just use value of Provisioner as `default` and not take the pain of setting it every time while running the acceptance test. As it will be the maintainers' responsibility to run them every time. Just a suggestion.

main.go
```diff
@@ -26,12 +24,7 @@ func main() {
 	}
 
 	if debugMode {
-		err := plugin.Debug(context.Background(), "vmware/tanzu/tmc", opts)
```
Vasundharashukla, line 29: This is changed due to the new linter version?

Vasundharashukla (changes requested):

shobha2626:
> Please make changes to the concerned flattening tests as well.

We are setting the flag deprecated in this release and will remove these fields completely in the next release. Till then we should provide the support

internal/resources/cluster/helper_test.go
shobha2626: What if we have different provisioners? Hence, it is better to keep it as an environment variable. We had a bug where we were setting the provisioner name to attach in case where we were only running the TKGm acceptance test, this fixes it.

main.go
shobha2626: No, this is part of Terraform SDK update. `Debug` function will soon be deprecated.

Vasundharashukla:
> > Please make changes to the concerned flattening tests as well.
> 
> We are setting the flag deprecated in this release and will remove these fields completely in the next release. Till then we should provide the support

Since this is not supported at the backend now, there is no point in allowing the users to set unnecessary fields. cc: @shreyassreenivas

shobha2626:
> > > Please make changes to the concerned flattening tests as well.
> > 
> > 
> > We are setting the flag deprecated in this release and will remove these fields completely in the next release. Till then we should provide the support
> 
> Since this is not supported at the backend now, there is no point in allowing the users to set unnecessary fields. cc: @shreyassreenivas

We can deprecate directly but it is always better to do it after setting it to deprecate

internal/client/errors/errors.go
```diff
@@ -16,12 +16,6 @@ type ClientErrors struct {
 	err      error
 }
 
-func Error(err error) ClientErrors {
```
shreyassreenivas, line 19: why is this removed ? 
due to unused ?

main.go
```diff
@@ -26,12 +24,7 @@ func main() {
...
-		if err != nil {
-			log.Fatal(err.Error())
-		}
-
-		return
+		opts.ProviderAddr = "vmware/dev/tanzu-mission-control"
```
shreyassreenivas, line 27: should we have dev ? what is the advantage of this ProviderAddr  ?

internal/resources/cluster/tkgvsphere/resource_tkgvsphere.go
```diff
@@ -576,12 +576,14 @@ var tkgVsphereNodePoolSpec = &schema.Schema{
...
 				Optional:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
+				Deprecated:  "node labels is not supported in TKGm vSphere",
 			},
 			cloudLabelKey: {
 				Type:        schema.TypeMap,
 				Description: "Cloud labels",
 				Optional:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
+				Deprecated:  "cloud labels is not supported in TKGm vSphere",
```
shreyassreenivas, line 586: deprecated is good, but 
is the main problem of the diff being reported addressed ?

internal/client/errors/errors.go
shobha2626: Yes, unused code

internal/resources/cluster/tkgvsphere/resource_tkgvsphere.go
shobha2626: Seems better to remove the field as backend is not supported

main.go
shobha2626: This is to test the provider changes locally in debug mode, since in the Makefile we build as `vmware/dev/tanzu-mission-control` using the same here

shreyassreenivas (dismissed):

shreyassreenivas:
Please get the PM approval before deleting the resource attributes

shreyassreenivas:
Squah the commits to one

Vasundharashukla (dismissed):

Vasundharashukla (approved):

shreyassreenivas:
@shobha2626 , can you please change the commit and the PR title as we are not deprecating these fields we are removing them

shreyassreenivas (approved):

PR #44, opened 2022-06-17
Update tanzu-mission-control_cluster.md

hveeran:
Removed `node_label` and `cloud_label`. These labels are removed since 1.0.2.

   Removed `node_label` and `cloud_label`. These labels are removed since 1.0.2.


shobha2626:
Hi @hveeran, thank you for your interest in TMC terraform provider project. We have incorporated the changes in this PR and have published as part of the patch release v1.0.4

PR #49, opened 2022-07-26
Handle manual delete of tf resources gracefully

shreyassreenivas:
2. 
When you create something in Terraform but delete it manually, Terraform
should gracefully handle it. If the API returns an error when the resource
doesn't exist, the read function should check to see if the resource is available
first. If the resource isn't available, the function should set the ID to an empty
string so Terraform "destroys" the resource in state

3. **Which issue(s) this PR fixes**

        Fixes #

4. **Additional information**

5. **Special notes for your reviewer**


shobha2626 (approved):
LGTM

ishangupta-ds (approved):

PR #50, opened 2022-08-02
Refactor policy resource service clients and models

ishangupta-ds:
**What this PR does / why we need it**: This PR refactors policy resource service clients and models.

PR #51, opened 2022-08-02
add request URL builder helper for cluster resource service client

ishangupta-ds:
Signed-off-by: Ishan Gupta <gishan@vmware.com>

**What this PR does / why we need it**: add request URL builder helper for cluster resource service client

Changes introduced as per: https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/50#discussion_r935436381

### DIFF
--- a/internal/client/cluster/cluster_resource.go
+++ b/internal/client/cluster/cluster_resource.go
@@ -6,11 +6,18 @@ SPDX-License-Identifier: MPL-2.0
 package clusterclient
 
 import (
-	"fmt"
 	"net/url"
 
 	"github.com/vmware-tanzu/terraform-provider-tanzu-mission-control/internal/client/transport"
+	"github.com/vmware-tanzu/terraform-provider-tanzu-mission-control/internal/helper"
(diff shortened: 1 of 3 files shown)


Vasundharashukla (dismissed):

internal/client/cluster/cluster_resource.go
```diff
@@ -1,18 +1,25 @@
 /*
-Copyright © 2021 VMware, Inc. All Rights Reserved.
+Copyright © 2022 VMware, Inc. All Rights Reserved.
```
shreyassreenivas, line 2: Do not change the copyrights, 
copyrights should only be added during the new file creation and not be update everytime the file gets modified

shreyassreenivas (changes requested):

internal/client/cluster/cluster_resource.go
ishangupta-ds: Sure, I will revert this change.

ishangupta-ds: Updated @shreyassreenivas

PR #71, opened 2022-09-27
 Add custom policy recipe to models

Vasundharashukla:
 Add custom policy recipe to models
 

3. **Additional information**
Added the models as per the new generation tool. Made some tweaks to the descriptions.


shreyassreenivas (approved):

internal/models/policy/recipe/custom/tmc_block_rolebinding_subjects.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"github.com/go-openapi/swag"
+
+	policyrecipecustomcommonmodel "github.com/vmware/terraform-provider-tanzu-mission-control/internal/models/policy/recipe/custom/common"
+)
+
+// VmwareTanzuManageV1alpha1CommonPolicySpecCustomV1TMCBlockRoleBindingSubjects Input schema for custom policy tmc-block-rolebinding-subjects recipe version v1.
+//
+// swagger:model vmware.tanzu.manage.v1alpha1.common.policy.spec.custom.v1.tmc.block.rolebinding.subjects
+type VmwareTanzuManageV1alpha1CommonPolicySpecCustomV1TMCBlockRoleBindingSubjects struct {
```
shreyassreenivas, line 19: why there are duplicate structs

ishangupta-ds (approved):

internal/models/policy/recipe/custom/test_host_networking.go
```diff
@@ -0,0 +1,90 @@
...
+
+	policyrecipecustomcommonmodel "github.com/vmware/terraform-provider-tanzu-mission-control/internal/models/policy/recipe/custom/common"
+)
+
+// VmwareTanzuManageV1alpha1CommonPolicySpecCustomV1TestHostNetworking test-host-networking recipe schema.
+//
+// The input schema for test-host-networking recipe.
+//
+// swagger:model VmwareTanzuManageV1alpha1CommonPolicySpecCustomV1TestHostNetworking
+type VmwareTanzuManageV1alpha1CommonPolicySpecCustomV1TestHostNetworking struct {
```
shreyassreenivas, line 20: should the file name start with test ? this is not test file right ?

internal/models/policy/recipe/custom/tmc_block_resources.go
```diff
@@ -0,0 +1,51 @@
...
+
+	policyrecipecustomcommonmodel "github.com/vmware/terraform-provider-tanzu-mission-control/internal/models/policy/recipe/custom/common"
+)
+
+// VmwareTanzuManageV1alpha1CommonPolicySpecCustomV1TMCBlockResources tmc-block-resources recipe schema.
+//
+// The input schema for tmc-block-resources recipe.
+//
+// swagger:model VmwareTanzuManageV1alpha1CommonPolicySpecCustomV1TMCBlockResources
+type VmwareTanzuManageV1alpha1CommonPolicySpecCustomV1TMCBlockResources struct {
```
shreyassreenivas, line 20: nodeport and block resources are same right ? what is the difference

shreyassreenivas:
If the struct has the same attributes cannot the same struct be reused instead of creating one for each type.? 
Can you explain the reason

internal/models/policy/recipe/custom/test_host_networking.go
Vasundharashukla: that's the name of the recipe `test-host-networking`

internal/models/policy/recipe/custom/tmc_block_resources.go
ishangupta-ds: Yeah since these are dependent on the gatekeeper CR spec which may be updated, thus we have kept the high level structs for the recipes different although they might appear similar for now.

Anything that is under one recipe kind and is repeated while generating models can be reused same applies for any non-primitive object model across different policies which is embedded inside the top level recipe structs.

internal/models/policy/recipe/custom/test_host_networking.go
Vasundharashukla: Realised that it was a user made recipe, hence removed it.

internal/models/policy/recipe/custom/tmc_block_resources.go
shreyassreenivas: I could not get what is the take away here. 
But I would suggest to avoid duplicating the structs if the same struct can we used again

ishangupta-ds (approved):

shreyassreenivas (approved):

PR #72, opened 2022-09-28
Use plugin version that matches tags

johnweldon:

    The `build` and `build-arm` targets for building local development versions were using the hard-coded version `1.0.0`

    This PR uses a generated version number, using the latest tag, thereby making it easier to keep multiple development versions of the plugin installed.

    With this fix, building the plugin locally on the current `main` branch would yield:

    `~/.terraform.d/plugins/vmware/dev/tanzu-mission-control/1.1.2/darwin_amd64/terraform-provider-tanzu-mission-control_v1.1.2`,

    instead of the hard-coded:

    `~/.terraform.d/plugins/vmware/dev/tanzu-mission-control/1.0.0/darwin_amd64/terraform-provider-tanzu-mission-control_v1.0.0`

    Additionally, instead of using two different make targets, the `build` target will place the generated binary in the directory that reflects the arch and os.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,17 +5,15 @@ ifeq ($(VERSION_TAG),)
 	VERSION_TAG := $(shell git describe --dirty --first-parent --always --tags)
 endif
 
+GOARCH := $(shell go env GOARCH)
+GOOS := $(shell go env GOOS)
+
 default: build
 
 build:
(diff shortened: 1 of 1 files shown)


shreyassreenivas (approved):
Thank you for the change. 
This is a clean solution

Vasundharashukla (approved):

PR #84, opened 2022-11-07
Add credential resource acceptance test

shreyassreenivas:

Add credential resource acceptance test

        Fixes #

3. **Additional information**


examples/resources/credential/resource.tf
```diff
@@ -0,0 +1,140 @@
+# Create AWS_EKS credential
```
Vasundharashukla, line 1: Can you please add the resource_template files as well. And rename the PR as Acceptance test with examples..

examples/resources/credential/resource.tf
```diff
@@ -0,0 +1,140 @@
...
+          arn = "arn:aws:iam::4987398738934:role/clusterlifecycle-test.tmc.cloud.vmware.com"
+          ext_id =""
+        }
+      }
+    }
+  }
+}
+
+
+# Create credential for TMC provisioned AWS S3 storage used for data-protection
```
Vasundharashukla, line 28: Let's split all these resource examples into different files, like we have done for cluster and policy resources for ease of user readability and documentation.

internal/resources/credential/credential_provider_test.go
```diff
@@ -0,0 +1,44 @@
...
+		ConfigureContextFunc: authctx.ProviderConfigureContext,
+	}
+	if err := testAccProvider.InternalValidate(); err != nil {
+		require.NoError(t, err)
+	}
+
+	return testAccProvider
+}
+
+const (
```
Vasundharashukla, line 38: Let's add the constants above the function.

Vasundharashukla (changes requested):

Vasundharashukla:
Please update the PR description as well, after making changes to the PR title.

Vasundharashukla:
Also, can we have attach the test run screenshot like we have done for the policies?

examples/resources/credential/resource.tf
shreyassreenivas: Resource_template file will be added during the documentation

Vasundharashukla: I mean we can do that way, but resource_templates are similar to examples, just that the values need to be replaced by placeholders. In documentation, we take up the templates and the generated docs, thoughts?

shreyassreenivas: The templating is better done during the Documentation task else going through the commit history might be confusing

examples/resources/credential/resource.tf
shreyassreenivas: Done

Vasundharashukla (approved):

PR #89, opened 2022-11-15
Add fix to acceptance tests for proxy support

Vasundharashukla:
Signed-off-by: Vasundhara Shukla <vasundharas@vmware.com>

        Fixing acceptance tests for proxy support.

### DIFF
--- a/internal/authctx/helper.go
+++ b/internal/authctx/helper.go
@@ -91,10 +91,10 @@ func ProviderConfigureContext(_ context.Context, d *schema.ResourceData) (interf
 	config.TLSConfig.ClientAuthKey, _ = d.Get(clientAuthKey).(string)
 	config.TLSConfig.CaCert, _ = d.Get(caCert).(string)
 
-	return setContext(config)
+	return setContext(&config)
 }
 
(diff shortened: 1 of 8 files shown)


ishangupta-ds (approved):

internal/authctx/helper.go
```diff
@@ -91,10 +91,10 @@ func ProviderConfigureContext(_ context.Context, d *schema.ResourceData) (interf
 	config.TLSConfig.ClientAuthKey, _ = d.Get(clientAuthKey).(string)
 	config.TLSConfig.CaCert, _ = d.Get(caCert).(string)
 
-	return setContext(config)
+	return setContext(&config)
 }
 
-func setContext(config TanzuContext) (TanzuContext, diag.Diagnostics) {
+func setContext(config *TanzuContext) (TanzuContext, diag.Diagnostics) {
```
shreyassreenivas, line 97: @ishangupta-ds , the proxy changes was manually tested right ?
why this change now?

ishangupta-ds: Yeah, the original one should also work as we are just initialising the field here but since for all other functions we have passed as pointer so for consistency we changed here as well.

shreyassreenivas (approved):

PR #94, opened 2022-11-18
EKS clusters - nodepool support

asutoshpalai:
Adds CRUD support for EKS Nodepool

3. **Additional information**
This is code complete but is marked as WIP because I haven't been able to test it yet.


internal/client/ekscluster/nodepool/nodepool_resource.go
```diff
@@ -0,0 +1,122 @@
+package eksnodepool
```
shreyassreenivas, line 1: copyright missing

internal/resources/ekscluster/data_source.go
```diff
@@ -71,16 +84,16 @@ func dataSourceTMCEKSClusterRead(ctx context.Context, d *schema.ResourceData, m
 	case "":
 		fallthrough
 	case "default":
-		timeoutValueData = "3m"
+		timeoutValueData = "30m"
```
shreyassreenivas, line 87: 30m is a long wait. 
can we reduce this to 5mins ?

internal/resources/ekscluster/resource_ekscluster.go
```diff
@@ -316,27 +325,211 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
 		return diag.FromErr(errors.Wrapf(err, "Unable to get Tanzu Mission Control EKS cluster entry, name : %s", d.Get(NameKey)))
 	}
 
+	opsRetryTimeout := getRetryTimeout(d)
+
 	clusterSpec := constructSpec(d)
 
 	// TODO: update nodepools seperatly
+	err = handleNodepoolDiffs(config, opsRetryTimeout, getResp.EksCluster.FullName, clusterSpec.NodePools)
+	if err != nil {
```
shreyassreenivas, line 334: Since the node pool and the cluster can be updated, in the provider we should also try doing the best possible. 
In this case, we should try updating nodepool and cluster, then return error if anyone of the operation failed.

Aborting the possible cluster update below is not an ideal approach

internal/resources/ekscluster/resource_ekscluster.go
```diff
@@ -316,27 +325,211 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
...
+	tmcCluster.Meta.CreationTime = strfmt.DateTime{}
 
-	_, err = config.TMCConnection.EKSClusterResourceService.EksClusterResourceServiceUpdate(
+	newCluster := &eksmodel.VmwareTanzuManageV1alpha1EksclusterEksCluster{
+		FullName: tmcCluster.FullName,
+		Meta:     tmcCluster.Meta,
+		Spec:     clusterSpec,
+	}
+
+	_, err := config.TMCConnection.EKSClusterResourceService.EksClusterResourceServiceUpdate(
```
shreyassreenivas, line 380: can you add a comment here explaining why cluster update does not take nodepool resource as well

internal/resources/ekscluster/resource_ekscluster.go
```diff
@@ -316,27 +325,211 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
...
+
+	// The TF update request was only for nodepools.
+	// No need to update Cluster, it will error out.
+	if !updateCluster {
+		return nil
+	}
+
+	// there is some translation error, which results
+	// in mismatch on the server.
+	tmcCluster.Meta.CreationTime = strfmt.DateTime{}
```
shreyassreenivas, line 372: ideally in the update, the meta attributes like creation time, uid etc should not be updated. 
even if the client provides an updated value, the changes should not be honoured

internal/resources/ekscluster/resource_ekscluster.go
```diff
@@ -316,27 +325,211 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
...
+		}
+	}
+
+	for _, tfNp := range nodepools {
+		if _, ok := tmcNps[tfNp.Info.Name]; !ok {
+			npCreate = append(npCreate, tfNp)
+		}
+	}
+
+	for _, np := range npCreate {
```
shreyassreenivas, line 430: suggestion, since we have dedicated function to handle create, update and delete of an nodepool. 
can the same function be enchanced to take the array and perform the required action

internal/models/ekscluster/nodepool_full_name.go
```diff
@@ -0,0 +1,48 @@
+package models
```
Vasundharashukla, line 1: Copyright information in all model files.

internal/resources/ekscluster/resource_ekscluster.go
```diff
@@ -77,7 +84,7 @@ var clusterSchema = map[string]*schema.Schema{
 var clusterSpecSchema = &schema.Schema{
 	Type:        schema.TypeList,
 	Description: "Spec for the cluster",
-	Required:    true,
+	Optional:    true,
```
Vasundharashukla, line 87: How come the spec of cluster is optional? Can you create the cluster without any spec fields?

internal/resources/ekscluster/data_source_test.go
```diff
@@ -0,0 +1,56 @@
+package ekscluster
```
shreyassreenivas, line 1: missing copyright

internal/resources/ekscluster/resource_ekscluster.go
asutoshpalai: As we are using the same Schema for data source and resource, I had to make the spec optional.

This is the same pattern as is followed in other resources. I can create a new schema if this is not appropriate.

internal/resources/ekscluster/resource_ekscluster.go
asutoshpalai: Yes, ideally we should. But the TMC API here checks if the update is set. If set, it checks that it is not modified. Without unsetting it, we get flaky errors when the conversion is not accurate and the time on the server doesn't match the time we pass.

internal/resources/ekscluster/data_source.go
```diff
@@ -109,11 +122,58 @@ func dataSourceTMCEKSClusterRead(ctx context.Context, d *schema.ResourceData, m
...
+		// check for deleted nodepools
+		for i := range nodepools {
+			if nodepools[i] == nil {
+				nodepools[i] = &eksmodel.VmwareTanzuManageV1alpha1EksclusterNodepoolDefinition{}
+			}
+		}
+
+		resp.EksCluster.Spec.NodePools = nodepools
+	} else {
+		resp.EksCluster.Spec.NodePools = clusterSpec.NodePools
```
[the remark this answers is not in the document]
asutoshpalai: Actually removed this code in refactoring.

shreyassreenivas (approved):

PR #95, opened 2022-11-20
Changing scope name fields for policy resource and correction of resource templates

Vasundharashukla:
Signed-off-by: Vasundhara Shukla <vasundharas@vmware.com>

- Change full name scope as per API naming for Policy resource
- Make changes to examples reflecting the same
- Rectify resource templates for the security policy resource


shreyassreenivas (dismissed):

internal/resources/policy/scope/constants.go
```diff
@@ -8,9 +8,9 @@ package scope
 const (
 	ManagementClusterNameKey = "management_cluster_name"
 	ProvisionerNameKey       = "provisioner_name"
-	ClusterNameKey           = "name"
+	ClusterNameKey           = "cluster_name"
```
shreyassreenivas, line 11: Since this resource has already been released, we should introduce new attributed and mark the old key as deprecated. 
Can you please do this changes to ensure backward compatibility

shreyassreenivas:
Can you please park this MR until we figure out how the upgrade scenario works out.

PR #117, opened 2023-01-27
Quota Policy Resource support for Terraform Provider

Vasundharashukla:
Support Quota Policy resource. Includes:

- Model for custom recipe
- Schema for all recipes, input and spec
- Flatten tests
- Resource implementation (CRUD)
- Acceptance tests
- Examples and resource templates

2. **Acceptance Test Success**:

Screenshot 2023-02-07 at 8 28 42 PM


go.mod
```diff
@@ -19,17 +17,7 @@ require (
...
-	github.com/PuerkitoBio/purell v1.1.1 // indirect
-	github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
-	github.com/go-openapi/analysis v0.19.5 // indirect
-	github.com/go-openapi/jsonpointer v0.19.3 // indirect
-	github.com/go-openapi/jsonreference v0.19.3 // indirect
-	github.com/go-openapi/loads v0.19.4 // indirect
-	github.com/go-openapi/spec v0.19.3 // indirect
-	github.com/google/go-cmp v0.5.8 // indirect
-)
+require github.com/google/go-cmp v0.5.8 // indirect
```
shreyassreenivas, line 20: can you fix this, why there is another block of require

internal/resources/policy/kind/quota/constants.go
```diff
@@ -0,0 +1,25 @@
...
+
+package policykindquota
+
+import (
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/policy"
+	reciperesource "github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/policy/kind/quota/recipe"
+)
+
+const (
+	ResourceName     = "tanzu-mission-control_namespace_quota_policy"
```
shreyassreenivas, line 14: why is this called namespace quota policy

shreyassreenivas:
Can you add the examples, it will be helpful in the reviews

internal/resources/policy/kind/quota/constants.go
Vasundharashukla: The name of the policy in API is set as type `namespace-quota-policy`

Screenshot 2023-02-03 at 3 10 00 PM

shobha2626: It's seems better to stick with just `quota-policy`?

internal/resources/policy/operations/helper.go
```diff
@@ -20,13 +20,15 @@ import (
 	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/policy"
 	policykindcustom "github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/policy/kind/custom"
 	policykindimage "github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/policy/kind/image"
+	policykindquota "github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/policy/kind/quota"
```
shobha2626, line 23: Why have repo path as policy/kind/...? Is the kind actually required?

internal/resources/policy/kind/quota/constants.go
Vasundharashukla: I followed the same convention as API, they are using namespace-quota-policy, even in policy-engine, they have named it as namespace-quota-policy where schemas have been described. For user interface, I have used `Quota policy` in documentation.

internal/resources/policy/operations/helper.go
Vasundharashukla: We are naming it as `policykind` as we have refactored operations and the type of policy in separate directories, so describing as kind made sense, let me know if that needs to be changed, but we will have to take it in a separate PR, as all previous policies will also required to be changed accordingly.

shreyassreenivas: Kind is not adding much value, it can be removed but at the sometime it's not too long, so for now we can live with this. 
If the import alias gets too big then consider removing it

internal/resources/policy/kind/quota/constants.go
shreyassreenivas: Is the quota policy is applied only for namespace level resources ? 

I feel just quota policy should be sufficient, but just check with Pranav once. 
What does the UI say ? follow the terminology used in UI. 
Like sourceSecret to repository_credentials

ishangupta-ds (approved):
LGTM.

internal/resources/policy/kind/quota/constants.go
Vasundharashukla: Quota policy is applied to org, cluster group and cluster level only. UI says Quota policy but request/ response from API has `namespace-quota-policy`. Have asked Pranav too.

shreyassreenivas: Then lets go with the UI and call it Quota policy

Vasundharashukla: Done

shreyassreenivas:
@Vasundharashukla , 
the commit message just says unit test and acceptance, this does not speak for which unit tests is being added

So always ensure that the commit message explain the changes.

internal/resources/common/objectmeta_schema.go
```diff
@@ -157,3 +157,14 @@ func GetTypeMapData(data map[string]interface{}) map[string]string {
 
 	return convertedMapData
 }
+
+func GetTypeIntMapData(data map[string]interface{}) map[string]int {
+	convertedMapData := make(map[string]int)
+
+	for key, value := range data {
+		value := value.(int)
```
shreyassreenivas, line 165: there is a possibility of code panic here

internal/resources/policy/kind/quota/constants.go
```diff
@@ -0,0 +1,25 @@
...
+package policykindquota
+
+import (
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/policy"
+	reciperesource "github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/policy/kind/quota/recipe"
+)
+
+const (
+	ResourceName     = "tanzu-mission-control_quota_policy"
+	typeDefaultValue = "namespace-quota-policy"
```
shreyassreenivas, line 15: Should this also be renamed into just quota_policy

internal/resources/policy/kind/quota/input_schema.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	RecipesAllowed = [...]string{reciperesource.CustomKey, reciperesource.SmallKey, reciperesource.MediumKey, reciperesource.LargeKey}
+)
+
+type (
+	Recipe string
+	// InputRecipe is a struct for all types of quota policy inputs.
+	inputRecipe struct {
+		recipe      Recipe
+		inputCustom *policyrecipequotamodel.VmwareTanzuManageV1alpha1CommonPolicySpecQuotaV1Custom
```
shreyassreenivas, line 45: why do you need two attributes here ? 
can't we derive from the recipe type  ?

internal/resources/policy/kind/quota/resource/resource_quota_policy_test.go
```diff
@@ -0,0 +1,396 @@
...
+				},
+				Config: testConfig.getTestQuotaPolicyResourceBasicConfigValue(scope.OrganizationScope, policykindquota.MediumRecipe),
+				Check:  testConfig.checkQuotaPolicyResourceAttributes(scope.OrganizationScope),
+			},
+		},
+	},
+	)
+
+	t.Log("quota policy resource acceptance test complete for medium recipe!")
+	time.Sleep(2 * time.Minute)
```
shreyassreenivas, line 168: why are these sleep statements needed ?

internal/resources/policy/kind/quota/constants.go
Vasundharashukla: We cannot do that as we get `400 Bad Request, response: {"error":"InvalidArgument: permission denied for policy type 'quota-policy'","code":3,"message":"InvalidArgument: permission denied for policy type 'quota-policy'` for this as in the API request, it accepts- `namespace-quota-policy`. Reference: screenshot in above comment. Only resource name can be changed, `typeDefaultValue` is mapped to  the Type in the API - `VmwareTanzuManageV1alpha1CommonPolicySpec` where Type is `Type of the policy object`.

internal/resources/policy/kind/quota/input_schema.go
Vasundharashukla: We have two different kinds of recipes (overall,4 - out of which 3 are the empty kinds with pre-defined values), for which we are using input<recipe> to define the spec. We have been using similar approach  in all policies, starting with Security policy as every recipe has a different set of attributes.

internal/resources/policy/kind/quota/resource/resource_quota_policy_test.go
Vasundharashukla: These ensure to give ample time to clean up the cluster/ cluster group created for a particular recipe type before starting the creation of another one to avoid conflict of helper resource- scope( cluster/ cluster group) and smooth running of acceptance tests.

Vasundharashukla: Again, this has been tested in the past for security and custom policy and since the issue was encountered, sleep was introduced in all the acceptance tests pertaining to policies where we have multiple recipes and multiple scopes unlike other resources.

internal/resources/policy/kind/quota/constants.go
shreyassreenivas: can you rename the variable so that the variable says this is being used in the API as policy type

internal/resources/policy/kind/quota/resource/resource_quota_policy_test.go
shreyassreenivas: Why can't we use new cluster and cluster group names ? 
having 2 mins sleep for every test will make the tests very slow and this does not guarantee that the cluster is actually cleaned. 
This is a wrong way of writing the acceptance test and rely on sleep is not accepted

internal/resources/policy/kind/quota/input_schema.go
shreyassreenivas: The third attribute is just being used for assigning nil and does not hold any values, so I believe this is something which can be avoided, Just give a thought

Vasundharashukla: This was the only way in which we could have defined the recipes with empty body and not by the recipe name to be in sync with other policies. As per our discussion in the past where I mentioned that we can just take recipe as string and pass nothing inside input, you suggested of not doing it and passing recipe inside input with nothing inside it. Please refer to the examples provided. If we have to remove this, we rather modify the whole quota policy and start give recipe name out of the input scope.

internal/resources/policy/kind/quota/resource/resource_quota_policy_test.go
Vasundharashukla: Can you suggest a better way to avoid sleep at all? We are using helper functions to create scopes, we would then need more of those. Again, this would then have to be changed for all other kinds of policies.

internal/resources/policy/kind/quota/input_schema.go
shreyassreenivas: I remember the discussion. 
The present schema defined looks good and would not need any change. 

This comment is based on the coding. 
Even if you have 4 recipes, 
The inputCustom and InputCommon are mutually exclusive to each other. 

If you see the values assigned. 
inputCommon is assigned as nil when InputCustom is not being used

internal/resources/policy/kind/quota/resource/resource_quota_policy_test.go
shreyassreenivas: I remember commenting on this earlier as well. 
There are two ways to address this. 
1. use new name for cluster and cluster group. This avoids the block on the clean up
2. come up with helper function which will wait only for the time duration needed for cluster clean up. 

I would prefer having option 1 implemented for all the acceptance test of policy. 

Their PR can be get merged but this needs to be fixed in the very next PR. Having 2 mins sleep is very costly for each acceptance test

@sbasu31

internal/resources/policy/kind/quota/constants.go
Vasundharashukla: I have replaced `typeDefaultValue` to `typePolicy` for all policies and added a comment as well.

internal/resources/policy/kind/quota/input_schema.go
ishangupta-ds: This would tie the logic very closely with the existing policy recipes. It's just a thought, but I think it would cause tight coupling.

I believe the current code is more readable and extensible. But feel free to update if you feel strongly about it.

ishangupta-ds (dismissed):
LGTM.

internal/resources/common/objectmeta_schema.go
```diff
@@ -157,3 +157,14 @@ func GetTypeMapData(data map[string]interface{}) map[string]string {
 
 	return convertedMapData
 }
+
+func GetTypeIntMapData(data map[string]interface{}) map[string]int {
```
shreyassreenivas, line 161: @ishangupta-ds @shobha2626 , this is where generics should be used

internal/resources/policy/kind/quota/input_schema.go
```diff
@@ -0,0 +1,181 @@
...
+				inputCustom: reciperesource.ConstructCustom(v1),
+			}
+		}
+	}
+
+	if v, ok := inputData[reciperesource.SmallKey]; ok {
+		if v1, ok := v.([]interface{}); ok && len(v1) != 0 {
+			inputRecipeData = &inputRecipe{
+				recipe:      SmallRecipe,
+				inputCommon: nil,
```
shreyassreenivas, line 70: explicit assignment of nil is not needed

internal/resources/policy/kind/quota/input_schema.go
Vasundharashukla: I have taken a single variable- `input` and used it both for custom and the pre-defined template recipes.

internal/resources/common/objectmeta_schema.go
ishangupta-ds: Should we take it up as an incremental change? Or is it required in this PR?

ishangupta-ds (approved):

ishangupta-ds (dismissed):

docs/resources/namespace_quota_policy.md
```diff
@@ -0,0 +1,743 @@
+---
+Title: "Quota Policy Resource"
+Description: |-
+    Creating the Tanzu Kubernetes quota policy resource.
+---
+
+# Quota Policy
```
shreyassreenivas, line 7: I believe all the title statements should also be renamed to namespace quota policy ?

Vasundharashukla: Done

shreyassreenivas (approved):

ishangupta-ds (approved):

PR #119, opened 2023-02-03
Force detach the clusters when the cluster is not connected to TMC

shreyassreenivas:

Force detach the clusters when the cluster is not connected to TMC

        Fixes #

3. **Additional information**

The changes has been validated manually by attaching and detaching the kind cluster. 
The SD ticket scenario was mimicked and the changes worked as expected 

### DIFF
--- a/internal/resources/credential/credential.go
+++ b/internal/resources/credential/credential.go
@@ -73,6 +73,7 @@ var dataSpec = &schema.Schema{
 	Optional:    true,
 	MaxItems:    1,
 	Description: "Holds credentials sensitive data",
+	Sensitive:   true,
 	Elem: &schema.Resource{
 		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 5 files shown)


ishangupta-ds (approved):

internal/resources/credential/credential.go
```diff
@@ -73,6 +73,7 @@ var dataSpec = &schema.Schema{
 	Optional:    true,
 	MaxItems:    1,
 	Description: "Holds credentials sensitive data",
+	Sensitive:   true,
```
Vasundharashukla, line 76: Should the docs be generated again for this addition in schema?

shreyassreenivas: Yes, docs should be generated, the generation of docs can be done when next doc update is scheduled.

Vasundharashukla (approved):

PR #121, opened 2023-02-09
Replacement of terraform resource naming to match Terraform conventions

VeeeneX:

Terraform uses `snake_case` to name the resources I stubled upon this while trying to access `data.resource-with-dashes`. 

None, but maybe also consider renaming the provider to use the underscores.

3. **Additional information**

- https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/naming#resource-names
- https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs (Also uses underscores)
- 

Please verify all the code, I haven't tested it just replaced a lot of files :)


shreyassreenivas:
@VeeeneX , Thank you for creating this PR. 

the ideology behind the provider naming is, 
<provider-name>_<resourcename>

In our case we had a choice of 

tanzumissioncontrol_cluster
or 
tanzu-mission-cluster_cluster

So more better readability, we have chosen `tanzu-mission-cluster_cluster`

Here, `tanzu-mission-cluster` is one product name hence we do not want to break it into multiple part with underscore.

VeeeneX:
@shreyassreenivas I don't see how `tanzu_mission_control_cluster` would break the product name it says Tanzu Mission Control Cluster and provider would be `tanzu_mission_control`

shreyassreenivas:
> @shreyassreenivas I don't see how `tanzu_mission_control_cluster` would break the product name it says Tanzu Mission Control Cluster and provider would be `tanzu_mission_control`

@VeeeneX , 
Terraform uses the first underscore to distinguish between the provider and the resource name
 
`tanzu_mission_control_cluster` would wrongly suggest that the `tanzu` is the provider and the resource name is `mission_control_cluster`

VeeeneX:
Ok, I see but still, those who are using know the name, changing between `-` and `_` is not really nice when it comes to referencing `data`

shreyassreenivas:
> Ok, I see but still, those who are using know the name, changing between `-` and `_` is not really nice when it comes to referencing `data`

Can you please explain this comment more ?

VeeeneX:
It's just this does not feel right :sweat_smile: `data.tanzu-mission-control_ekscluster.main` this mix of dashes and underscored

shreyassreenivas:
> It's just this does not feel right 😅 `data.tanzu-mission-control_ekscluster.main` this mix of dashes and underscored

Oh, I see. 
We had to take a call between the readability and dropping `-`. The team voted for `tanzu-mission-control`. 
Will probably revisit this when we are doing a major upgrade of the provider.

shreyassreenivas:
Closing the PR based on the above discussions

PR #125, opened 2023-02-21
Support raw kubeconfig input for attach clusters as an alternative input method

### LINKED ISSUE #77: kubeconfig input alternate options
### Is your feature request related to a problem? Please describe.

When using the attach_cluster_with_kubeconfig you must specify a path to the kubeconfig using the kubeconfig_file variable and can't use the content of an output or block as an input to the config needed.  There are extra modules needed to create a file in order to do this and creates more dependencies.

### Describe the solution you'd like

For example, when attaching an AKS cluster, it would be nice to use the [kube_config_raw](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/data-sources/kubernetes_cluster#kube_config_raw) or kube_config_admin_raw value as an input for the kubeconfig needed to attach a cluster to TMC.  Currently for an AKS attach, you must create the file through another method like local_file from the kube_config_raw variable and pass that file location to the TMC module.  So in the function of attach_k8s_cluster maybe have another input option that could be the content of another ouput.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

Vasundharashukla:
Add kubeconfig as raw input to the terraform script as an alternative option along with passing the file address for attach cluster resource.

        Fixes #77 

3. **Cluster attach verification**

Screenshot 2023-02-21 at 4 24 15 PM

Screenshot 2023-02-21 at 4 25 26 PM


internal/resources/cluster/resource_cluster.go
```diff
@@ -218,15 +227,27 @@ func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, m interf
...
 	var (
 		kubeconfigfile string
+		rawkubeconfig  string
 		k8sclient      k8sClient.Client
 		err            error
 	)
 
 	if _, ok := d.GetOk(attachClusterKey); ok {
-		kubeconfigfile, _ = d.Get(helper.GetFirstElementOf(attachClusterKey, attachClusterKubeConfigKey)).(string)
+		if v, ok := d.GetOk(helper.GetFirstElementOf(attachClusterKey, attachClusterKubeConfigPathKey)); ok {
```
shreyassreenivas, line 236: can you use switch here to construct the k8s client from file or raw config

internal/resources/cluster/resource_cluster.go
```diff
@@ -250,10 +271,10 @@ func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, m interf
 	// always run
 	d.SetId(clusterResponse.Cluster.Meta.UID)
 
-	if kubeconfigfile != "" {
+	if _, ok := d.GetOk(attachClusterKey); ok {
```
shreyassreenivas, line 274: This condition functionality will fail when attachClusterKey is defined but none of the attributes are added inside the block

internal/resources/cluster/resource_cluster.go
```diff
@@ -425,18 +446,71 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
...
 }
 
-func getK8sClient(kubeConfigFile string) (k8sClient.Client, error) {
-	restConfig, err := clientcmd.BuildConfigFromFlags("", kubeConfigFile)
-	if err != nil {
-		return nil, errors.WithMessagef(err, "Invalid kubeconfig file path provided, filepath : %s", kubeConfigFile)
+type (
+	KubeConfigOption func(*KubeConfig)
+
+	KubeConfigSpec struct {
```
shreyassreenivas, line 452: This change should be in next MR and it is not related to the present change

internal/resources/cluster/resource_cluster.go
```diff
@@ -425,18 +446,71 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
...
+	}
+
+	KubeConfig struct {
+		Path string
+		Raw  string
+		Spec *KubeConfigSpec
+	}
+)
+
+func WithPath(p string) KubeConfigOption {
```
shreyassreenivas, line 466: Good use of functional arguments

internal/resources/cluster/resource_cluster.go
```diff
@@ -425,18 +446,71 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
...
+	for _, o := range opts {
+		o(cfg)
+	}
+
+	var (
+		restConfig *rest.Config
+		err        error
+	)
+
+	if cfg.Path != "" {
```
shreyassreenivas, line 490: use switch statement here

internal/resources/cluster/resource_cluster.go
Vasundharashukla: For that terraform will error out from schema as expected min items is 1.

shreyassreenivas: That block has description , so if you just provide description and not other values how will ur code work ?

Vasundharashukla: Is the description really needed as a parameter to be passed, I just see it defined in the schema and not used any where else in the code base, is it something that will be largely used by the provider users?

shreyassreenivas: I am not talking about the use case here.
Pointing to the the code and the corner case which you are missing

shreyassreenivas: Description inside the attach k8s might be used by the user to describe something which would help them identify the kubeconfig which is being used.

internal/resources/cluster/resource_cluster.go
Vasundharashukla: Done

internal/resources/cluster/resource_cluster.go
Vasundharashukla: Added schema validation which will ensure that one of the ways of defining kubeconfig have been added and it will throw validation error from terraform to handle this case.

internal/resources/cluster/resource_cluster.go
Vasundharashukla: Done

docs/data-sources/cluster.md
```diff
@@ -80,7 +80,7 @@ data "tanzu-mission-control_cluster" "read_tkg_aws_cluster" {
 
 ### Optional
 
-- `attach_k8s_cluster` (Block List, Max: 1) (see [below for nested schema](#nestedblock--attach_k8s_cluster))
+- `attach_k8s_cluster` (Block List, Max: 2) (see [below for nested schema](#nestedblock--attach_k8s_cluster))
```
shreyassreenivas, line 83: why is this max two ?

internal/resources/cluster/resource_cluster.go
```diff
@@ -27,6 +31,12 @@ import (
 	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/common"
 )
 
+const (
+	unknown = iota
```
shreyassreenivas, line 35: is this really needed ? can this be avoided ?

internal/resources/cluster/resource_cluster.go
```diff
@@ -39,6 +49,7 @@ func ResourceTMCCluster() *schema.Resource {
 		UpdateContext: resourceClusterInPlaceUpdate,
 		DeleteContext: resourceClusterDelete,
 		Schema:        clusterSchema,
+		CustomizeDiff: ValidateKubeConfig,
```
shreyassreenivas, line 52: adding schema validation in customiseDiff is wrong way. 
From what I understand, customiseDiff is used to show the diffs and override the diffs shown or not.
Can you check the function definition in framework ? I might be wrong here

internal/resources/cluster/resource_cluster.go
```diff
@@ -94,26 +105,40 @@ func constructFullname(d *schema.ResourceData) (fullname *clustermodel.VmwareTan
...
-			attachClusterDescriptionKey: {
-				Type:        schema.TypeString,
-				Description: "Attach cluster description",
-				Optional:    true,
+var (
+	attachCluster = &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
+		MinItems: 1,
+		MaxItems: 2,
```
shreyassreenivas, line 113: why should the max items be 2

internal/resources/cluster/resource_cluster.go
```diff
@@ -213,20 +238,70 @@ func flattenSpec(spec *clustermodel.VmwareTanzuManageV1alpha1ClusterSpec) (data
 	return []interface{}{flattenSpecData}
 }
 
+func ValidateKubeConfig(ctx context.Context, diff *schema.ResourceDiff, i interface{}) error {
```
shreyassreenivas, line 241: The kubeconfig can be validated by checking if the k8s client is nil or not

internal/resources/cluster/resource_cluster.go
shreyassreenivas: Here, we should be checking is k8s client is nil or not

docs/data-sources/cluster.md
Vasundharashukla: One element is description, one is the kube_config input (either of the 2 available inputs), so that user doesn't provide multiple inputs.

internal/resources/cluster/resource_cluster.go
Vasundharashukla: For the switch case you asked to use instead of if conditions.

internal/resources/cluster/resource_cluster.go
Vasundharashukla: Same reason as above

internal/resources/cluster/resource_cluster.go
Vasundharashukla: A schema validation checks before even the call is made to TMC, so one the first layer we get the response.

docs/data-sources/cluster.md
shreyassreenivas: this limit is for the entries in they list. 
kuebconfig and description together will be on entry.

internal/resources/cluster/resource_cluster.go
shreyassreenivas: Switch can be implemented without this iota and this is not needed

internal/resources/cluster/resource_cluster.go
shreyassreenivas: you can use : ValidateFunc or ExactlyOneOf  []string attributes from the schema. 
We should not use diff customise diff to handle schema validations

internal/resources/cluster/resource_cluster.go
Vasundharashukla: Can you suggest the switching parameter passing the input kubeconfig value to k8sclient?

internal/resources/cluster/resource_cluster.go
Vasundharashukla: I believe than we need to change the same for other resources as well, customise diff has been used in other resources as well

internal/resources/cluster/resource_cluster.go
shreyassreenivas: you can use 
switch true{

}


or just switch {

}

This will solve ur use case

internal/resources/cluster/resource_cluster.go
shreyassreenivas: if it is only limited to schema validation and something which can be handled via validateFunc we should be using it. 
if other resources needs change, create a task and it can be picked in the future

internal/resources/cluster/resource_cluster.go
Vasundharashukla: Ref- https://github.com/hashicorp/terraform/pull/14887#issue-231839744

Usecase of `CustomizeDiff` from the terraform org: 

> Vetoing changes to the diff that could not be reasonably carried out (ie: illegal AWS RDS operations) by kicking back an error to the CustomizeDiff function, much like the other CRUD functions.

internal/resources/cluster/resource_cluster.go
Vasundharashukla: Added the check

docs/data-sources/cluster.md
Vasundharashukla: Done

internal/resources/cluster/resource_cluster.go
Vasundharashukla: Removed it and added switch case

internal/resources/cluster/resource_cluster.go
Vasundharashukla: As per the above use case, we can use `CustomizeDiff` for validations.

internal/resources/cluster/resource_cluster.go
Vasundharashukla: `validateFunc` is being used to check non empty Kube_config inputs. Here we are checking some extra corner cases such as not giving more than one Kube_config input type

ishangupta-ds:
Did we add the check for 'ready_wait_timeout', this should be optional if the customer does not want to attach the cluster right?

Cc: @shreyassreenivas

docs/data-sources/cluster.md
```diff
@@ -98,7 +98,8 @@ data "tanzu-mission-control_cluster" "read_tkg_aws_cluster" {
 Optional:
 
 - `description` (String) Attach cluster description
-- `kubeconfig_file` (String) Attach cluster KUBECONFIG path
+- `kube_config_path` (String) Attach cluster KUBECONFIG path
```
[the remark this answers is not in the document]
Vasundharashukla: Reversed the change

internal/resources/cluster/resource_cluster.go
```diff
@@ -213,20 +232,84 @@ func flattenSpec(spec *clustermodel.VmwareTanzuManageV1alpha1ClusterSpec) (data
...
+		}
+
+		isKubeConfigPresent := func(typeKey string) bool {
+			if value, ok := d.GetOk(helper.GetFirstElementOf(attachClusterKey, typeKey)); ok {
+				if value != nil {
+					kubeConfig = value
+					return true
+				}
+
+				return false
```
shreyassreenivas, line 293: This return is duplicate

internal/resources/cluster/resource_cluster.go
```diff
@@ -274,6 +357,66 @@ func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, m interf
...
+		}
+	}
+
+	if restConfig == nil {
+		return nil, errors.WithMessagef(err, "Kubeconfig not provided.")
+	}
+
+	restConfig.Timeout = 10 * time.Second
+
+	client, err := k8sClient.New(restConfig, k8sClient.Options{})
```
shreyassreenivas, line 412: check what is the value of the client in the case of the error and you can use it the above if condition check

shreyassreenivas (changes requested):

internal/resources/cluster/resource_cluster.go
Vasundharashukla: Added

ishangupta-ds (dismissed):

internal/resources/cluster/resource_cluster.go
```diff
@@ -94,26 +99,40 @@ func constructFullname(d *schema.ResourceData) (fullname *clustermodel.VmwareTan
...
+					Optional:     true,
+					ForceNew:     true,
+					Sensitive:    true,
+					ValidateFunc: validation.StringIsNotWhiteSpace,
+				},
+				attachClusterDescriptionKey: {
+					Type:         schema.TypeString,
+					Description:  "Attach cluster description",
+					Optional:     true,
+					ValidateFunc: validation.StringIsNotWhiteSpace,
```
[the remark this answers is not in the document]
Vasundharashukla: Yes, tested. Refer to the comment [here](https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/125#discussion_r1121620614)

internal/resources/cluster/resource_cluster.go
```diff
@@ -274,6 +357,66 @@ func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, m interf
 	return append(diags, dataSourceTMCClusterRead(context.WithValue(ctx, contextMethodKey{}, "create"), d, m)...)
 }
 
+type (
+	KubeConfigOption func(*KubeConfig)
```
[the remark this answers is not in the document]
Vasundharashukla: made all variables local

internal/resources/cluster/resource_cluster.go
```diff
@@ -274,6 +357,66 @@ func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, m interf
...
+
+	KubeConfig struct {
+		Path string
```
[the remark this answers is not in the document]
Vasundharashukla: done

internal/resources/cluster/resource_cluster.go
Vasundharashukla: removed

internal/resources/cluster/resource_cluster.go
Vasundharashukla: added the check above

docs/resources/cluster.md
```diff
@@ -42,6 +42,7 @@ resource "tanzu-mission-control_cluster" "attach_cluster_without_apply" {
     cluster_group = "default" # Default: default
   }
 
+  ready_wait_timeout = "0s"
```
shreyassreenivas, line 45: is this intended change ?

docs/resources/cluster.md
```diff
@@ -1029,6 +1061,11 @@ Optional:
 - `disk_size` (String) Root disk size in gigabytes for the VM
 - `memory` (String) Memory associated with the node in megabytes
 
+
```
shreyassreenivas, line 1064: why white spaces ?

examples/resources/cluster/resource_attach_cluster.tf
```diff
@@ -13,5 +13,6 @@ resource "tanzu-mission-control_cluster" "attach_cluster_without_apply" {
     cluster_group = "default" # Default: default
   }
 
+  ready_wait_timeout = "0s"
```
shreyassreenivas, line 16: can you explain why is this needed ?

internal/resources/cluster/resource_cluster.go
```diff
@@ -213,24 +232,87 @@ func flattenSpec(spec *clustermodel.VmwareTanzuManageV1alpha1ClusterSpec) (data
...
+			k8sclient, err = getK8sClient(withRaw(rawKubeConfig))
+		}
+
+		if err != nil {
+			log.Println("[ERROR] error while creating kubernetes client: ", err.Error())
+			return diag.FromErr(err)
+		}
+
+		if k8sclient == nil {
+			err = errors.New("k8s client not found")
```
shreyassreenivas, line 312: check the error message, k8s client is something we are creating, so not found will be a wrong error

internal/resources/cluster/resource_cluster.go
```diff
@@ -263,7 +345,7 @@ func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, m interf
 
 		log.Printf("[INFO] Applying %s cluster's deployment link manifest objects on to kubernetes cluster", constructFullname(d).ToString())
 
-		err = manifest.Create(k8sclient, deploymentManifests, true)
+		err = manifest.Create(*k8sclient, deploymentManifests, true)
```
shreyassreenivas, line 348: modify the manifest.Create function parameter to accept pointer k8sclient

examples/resources/cluster/resource_attach_cluster.tf
ishangupta-ds: @shreyassreenivas I asked for this change, basically if we are not attaching the cluster and only creating it (without the attach block) then I don't see any reason to block the terraform apply for 3 minutes.

WDYT

docs/resources/cluster.md
ishangupta-ds: Yes

docs/data-sources/cluster.md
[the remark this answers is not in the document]
Vasundharashukla: that's autogenerated from the tool, not something added by the developer.

docs/data-sources/cluster.md
```diff
@@ -99,6 +99,7 @@ Optional:
 
 - `description` (String) Attach cluster description
 - `kubeconfig_file` (String) Attach cluster KUBECONFIG path
+- `kube_config_raw` (String, Sensitive) Attach cluster KUBECONFIG
```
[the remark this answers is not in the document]
Vasundharashukla: yes, that's why for cluster scenario without kubeconfig, it doesn't make sense to wait for even 3 minutes. So, documentation can support this case.

examples/resources/cluster/resource_attach_cluster.tf
Vasundharashukla: Added to the documentation, so that users can take reference and not wait for 3m when kubeconfig is not present.

internal/resources/cluster/resource_cluster.go
Vasundharashukla: changed

internal/resources/cluster/resource_cluster.go
Vasundharashukla: done

internal/resources/cluster/resource_cluster.go
```diff
@@ -274,6 +356,66 @@ func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, m interf
...
+	}
+}
+
+func withRaw(r string) kubeConfigOption {
+	return func(config *kubeConfig) {
+		config.raw = r
+	}
+}
+
+func getK8sClient(opts ...kubeConfigOption) (*k8sClient.Client, error) {
```
shreyassreenivas, line 380: would it make sense to move this function inside the package where the k8s client is being consumed ie manifest package.

If you are moving, move both the function and the type variables

internal/resources/cluster/manifest/manifest.go
```diff
@@ -31,7 +31,7 @@ func Create(
 		return errors.WithMessage(err, "failure to fetch attach manifests")
 	}
 
-	toBeCleaned, err := objectsToBeCleaned(k8sclient, manifests, forceClean)
+	toBeCleaned, err := objectsToBeCleaned(*k8sclient, manifests, forceClean)
```
shreyassreenivas, line 34: objectsToBeCleaned(*k8sclient, manifests, forceClean)

*k8sclient, looks odd and should be avoided, so can you change the function definition for objectsToBeCleaned and 
createObjects

internal/resources/cluster/resource_cluster.go
Vasundharashukla: Then we would need to export all variables and functions by making them global. Since, only cluster resource is using this, we can have it here.

ishangupta-ds (approved):

shreyassreenivas (approved):

PR #129, opened 2023-02-27
refresh auth token silently when token expires

kduan005:
Refresh auth token in the background when it expires.

        Fixes #OLYMP-36711

3. **Additional information**
Screen Shot 2023-02-28 at 5 43 13 PM

### DIFF
--- a/internal/authctx/authentication.go
+++ b/internal/authctx/authentication.go
@@ -134,3 +134,10 @@ func getUserAuthCtx(config *TanzuContext) (map[string]string, error) {
 
 	return md, nil
 }
+
+func RefreshUserAuthCtx(config *TanzuContext) {
+	md, _ := getUserAuthCtx(config)
+	for key, value := range md {
+		config.TMCConnection.Headers.Set(key, value)
+	}
+}
--- a/internal/client/errors/errors.go
+++ b/internal/client/errors/errors.go
@@ -41,6 +41,24 @@ func IsNotFoundError(err error) bool {
 	return false
 }
 
+func IsUnauthorizedError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	convertedError, ok := err.(ClientErrors)
+	if !ok {
+		return strings.Contains(err.Error(), fmt.Sprintf("%d", http.StatusUnauthorized)) &&
+			strings.Contains(err.Error(), http.StatusText(http.StatusUnauthorized))
+	}
+
+	if convertedError.httpCode == http.StatusUnauthorized {
+		return true
+	}
+
+	return false
+}
+
 func (e ClientErrors) Error() string {
 	if e.err == nil {
 		return ""
--- a/internal/resources/cluster/data_source_cluster.go
+++ b/internal/resources/cluster/data_source_cluster.go
@@ -40,6 +40,12 @@ func dataSourceTMCClusterRead(ctx context.Context, d *schema.ResourceData, m int
 		err   error
 	)
 
+	refreshUserAUthCtx := func(config *authctx.TanzuContext, refreshCondition func(error) bool) {
+		if refreshCondition(err) {
+			authctx.RefreshUserAuthCtx(config)
+		}
+	}
+
 	getClusterResourceRetryableFn := func() (retry bool, err error) {
 		resp, err = config.TMCConnection.ClusterResourceService.ManageV1alpha1ClusterResourceServiceGet(constructFullname(d))
 		if err != nil {
@@ -47,6 +53,9 @@ func dataSourceTMCClusterRead(ctx context.Context, d *schema.ResourceData, m int
 				_ = schema.RemoveFromState(d, m)
 				return false, nil
 			}
+
+			// refresh auth bearer token if it expired
+			refreshUserAUthCtx(&config, clienterrors.IsUnauthorizedError)
 
 			return true, errors.Wrapf(err, "Unable to get Tanzu Mission Control cluster entry, name : %s", d.Get(NameKey))
 		}


internal/resources/cluster/data_source_cluster.go
```diff
@@ -47,6 +47,9 @@ func dataSourceTMCClusterRead(ctx context.Context, d *schema.ResourceData, m int
 				_ = schema.RemoveFromState(d, m)
 				return false, nil
 			}
+			if clienterrors.IsUnauthorizedError(err) {
```
shreyassreenivas, line 50: there should be refresh here

internal/authctx/client.go
```diff
@@ -38,26 +38,28 @@ func (cfg *TanzuContext) Setup() (err error) {
...
 	cfg.TMCConnection = client.NewTestHTTPClientWithDefaultTransport()
-	return setup(cfg)
+	cfg.TMCConnection.WithHost(cfg.ServerEndpoint)
+	cfg.TMCConnection.Headers.Set("Host", cfg.ServerEndpoint)
+
+	return SetAuthCtx(cfg)
 }
 
-func setup(cfg *TanzuContext) (err error) {
+func SetAuthCtx(cfg *TanzuContext) (err error) {
```
shreyassreenivas, line 57: why is this change needed ?

internal/authctx/client.go
```diff
@@ -38,26 +38,28 @@ func (cfg *TanzuContext) Setup() (err error) {
 	if err != nil {
 		return err
 	}
+	cfg.TMCConnection.WithHost(cfg.ServerEndpoint)
```
shreyassreenivas, line 41: these changes are not needed, can you please explain why this change needed ?

internal/resources/cluster/data_source_cluster.go
kduan005: Yeah so instead of creating a new fresh util function, I'm trying to reuse the `setup()` function in `client.go` from authctx package, because essentially the setup() does two things:
- get a new auth token and set the value in `Client.Header` (which can be reused for refreshing tokens)
- set host value in `Client.Header`

And that's why I moved the second point out to where `setup()` are called, so that we don't set host value with redundancy. Feel free to let me know if a separate refresh util function is preferred.

internal/authctx/client.go
kduan005: Same as explained in below, trying to move out setting host in headers to where `setup()` is called, so that we can reuse `setup()` for refreshing auth tokens

internal/authctx/client.go
kduan005: Trying to rename it so that we can reuse it elsewhere

internal/resources/cluster/data_source_cluster.go
kduan005: Okay so just made an update, and instead of reusing `setup()`, changed it to reusing `getUserAuthCtx()` for refreshing tokens

internal/resources/cluster/data_source_cluster.go
```diff
@@ -48,6 +48,13 @@ func dataSourceTMCClusterRead(ctx context.Context, d *schema.ResourceData, m int
 				return false, nil
 			}
 
+			if clienterrors.IsUnauthorizedError(err) {
+				md, _ := authctx.GetUserAuthCtx(&config)
```
shreyassreenivas, line 52: Can we have a function defined in the terms of `refreshUserAuthCtx` and this function has these lines 

Func RefreshUserAuthCtx(){
md, _ := authctx.GetUserAuthCtx(&config)
 				for key, value := range md {
 					config.TMCConnection.Headers.Set(key, value)
 				}
}

And lets not expose GetUserAuthCtx

shreyassreenivas: infact you can define functional parameters to enhance the refresh usecase. 

RefreshUserAuthCtx(withRefreshOnAuthError(err))
RefreshUserAuthCtx(withForceRefresh())

This would mean that the error check is also pushed into the refresh logic

shreyassreenivas:
Please validate if the change is working by manually testing the changes 
Test scenario
Use a attach cluster use case
set the wait period of 45minutes
Once the attach is successfully, delete the extension pod, before it communicates to tmc
This is force the cluster to wait until it is in ready state

internal/resources/cluster/data_source_cluster.go
kduan005: I've updated the code according to the comment, PTAL. I'll try to get Vasundhara or Shobha's help on the test, as right now the acceptance test seems to detach the cluster prematurely before the cluster come up to the ready state. I'll post my update once I complete the test.

kduan005:
> Please validate if the change is working by manually testing the changes Test scenario Use a attach cluster use case set the wait period of 45minutes Once the attach is successfully, delete the extension pod, before it communicates to tmc This is force the cluster to wait until it is in ready state

Screen Shot 2023-03-02 at 4 07 09 PM

I've tested with a kind cluster and added some extra logs in my local to show that the token was refreshed during the wait time, as seen in the above screenshot.

internal/authctx/authentication.go
```diff
@@ -134,3 +134,12 @@ func getUserAuthCtx(config *TanzuContext) (map[string]string, error) {
 
 	return md, nil
 }
+
+func RefreshUserAuthCtx(config *TanzuContext, refreshCondition func(error) bool, err error) {
```
shreyassreenivas, line 138: why is error needed to be returned?

internal/client/errors/errors.go
```diff
@@ -41,6 +41,24 @@ func IsNotFoundError(err error) bool {
 	return false
 }
 
+func IsUnauthorizedError(err error) bool {
```
shreyassreenivas, line 44: can you modify this function to retrun another function which will be invoked in the refreshAuthxCtx

shreyassreenivas: This will reduce the error being sent as a parameter to refreshAuthxCtx

shreyassreenivas (approved):

Vasundharashukla (approved):

PR #130, opened 2023-03-01
Making control plane endpoint optional for TKG vSphere cluster

Vasundharashukla:
vSphere controlplane endpoint ip should not be required when enabled AVI when you create a legacy cluster on TKGm. Making it optional.


shreyassreenivas (approved):

shreyassreenivas:
Can you add an example where AVI is used

ishangupta-ds (approved):

PR #135, opened 2023-03-09
Handle case of not found cluster entry in Cluster data source

Vasundharashukla:
The provider behaved abruptly and crashed when the data source was called on a non-existent cluster, to avoid the crush the error needed to be handled.

        Fixes #

3. **Additional information**
The PR throws the right error for an invalid get call and avoids provider crash. Moreover, the documentation has been changed to show that data source needs to be called for a valid (existing) Tanzu Mission Control resource.

4. **Error handling proof:**

Screenshot 2023-03-09 at 1 52 53 PM

### DIFF
--- a/internal/resources/cluster/data_source_cluster.go
+++ b/internal/resources/cluster/data_source_cluster.go
@@ -25,12 +25,14 @@ import (
 
 func DataSourceTMCCluster() *schema.Resource {
 	return &schema.Resource{
-		ReadContext: dataSourceTMCClusterRead,
-		Schema:      clusterSchema,
+		ReadContext: func(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+			return dataSourceClusterRead(context.WithValue(ctx, contextMethodKey{}, "dataSourceRead"), d, m)
+		},
+		Schema: clusterSchema,
 	}
 }
 
-func dataSourceTMCClusterRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+func dataSourceClusterRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 	config := m.(authctx.TanzuContext)
 
 	// Warning or errors can be collected in a slice type
@@ -50,7 +52,12 @@ func dataSourceTMCClusterRead(ctx context.Context, d *schema.ResourceData, m int
 		resp, err = config.TMCConnection.ClusterResourceService.ManageV1alpha1ClusterResourceServiceGet(constructFullname(d))
 		if err != nil {
 			if clienterrors.IsNotFoundError(err) {
+				if ctx.Value(contextMethodKey{}) == "dataSourceRead" {
+					return false, errors.Wrapf(err, "Tanzu Mission Control cluster entry not found, name : %s", d.Get(NameKey))
+				}
+
 				_ = schema.RemoveFromState(d, m)
+
 				return false, nil
 			}
 
@@ -105,7 +112,7 @@ func dataSourceTMCClusterRead(ctx context.Context, d *schema.ResourceData, m int
 		_, err = helper.RetryUntilTimeout(getClusterResourceRetryableFn, 10*time.Second, timeoutDuration)
 	}
 
-	if err != nil || resp == nil {
+	if err != nil || resp == nil || resp.Cluster == nil {
 		return diag.FromErr(errors.Wrapf(err, "Unable to get Tanzu Mission Control cluster entry, name : %s", d.Get(NameKey)))
 	}
 
--- a/docs/data-sources/cluster.md
+++ b/docs/data-sources/cluster.md
@@ -42,28 +42,28 @@ You must also have the appropriate permissions:
 ## Example Usage
 
 ```terraform
-# Read Tanzu Mission Control cluster : fetch cluster details
+# Read Tanzu Mission Control cluster : fetch cluster details for an already attached cluster
 data "tanzu-mission-control_cluster" "ready_only_attach_cluster_view" {
   management_cluster_name = "attached"       # Default: attached
   provisioner_name        = "attached"       # Default: attached
   name                    = "terraform-test" # Required
 }
 
-# Read Tanzu Mission Control Tanzu Kubernetes Grid Service workload cluster : fetch TKGs cluster details
+# Read Tanzu Mission Control Tanzu Kubernetes Grid Service workload cluster : fetch cluster details for already present TKGs cluster
 data "tanzu-mission-control_cluster" "read_tkgs_cluster" {
   management_cluster_name = "test-tkgs"
   provisioner_name        = "test-gc-e2e-demo-ns"
   name                    = "tkgs-workload"
 }
 
-# Read Tanzu Mission Control Tanzu Kubernetes Grid vSphere workload cluster : fetch TKG vSphere cluster details
+# Read Tanzu Mission Control Tanzu Kubernetes Grid vSphere workload cluster : fetch cluster details for already present TKG vSphere cluster
 data "tanzu-mission-control_cluster" "read_tkg_vsphere_cluster" {
   management_cluster_name = "tkgm-vsphere"
   provisioner_name        = "default"
   name                    = "tkgm-workload"
 }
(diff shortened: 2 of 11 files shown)


ishangupta-ds (approved):

internal/resources/cluster/data_source_cluster.go
```diff
@@ -45,7 +45,7 @@ func dataSourceTMCClusterRead(ctx context.Context, d *schema.ResourceData, m int
 		if err != nil {
 			if clienterrors.IsNotFoundError(err) {
 				_ = schema.RemoveFromState(d, m)
-				return false, nil
+				return false, errors.Wrapf(err, "Tanzu Mission Control cluster entry not found, name : %s", d.Get(NameKey))
```
shreyassreenivas, line 48: we should not return this error i guess

Vasundharashukla: The plugin crashes otherwise, at this code point the error should be returned in case of cluster not found. 
If we added the condition of early return without the error, then only terraform emitted warnings, that to which are only catched if running provider in DEBUG mode. 

The warnings- 

`2023-03-09T13:15:55.888+0530 [WARN]  Provider "provider[\"vmware/dev/tanzu-mission-control\"]" produced an unexpected new value for data.tanzu-mission-control_cluster.tf-cluster.
      - .management_cluster_name: was cty.StringVal("vasundharas-tkgs"), but now null
      - .name: was cty.StringVal("tf-cluster-new-2"), but now null
      - .provisioner_name: was cty.StringVal("testns"), but now null
`

shreyassreenivas: and this data source handler function is also invoked by resource, so check for the possible corner cases

internal/resources/cluster/data_source_cluster.go
```diff
@@ -25,116 +25,162 @@ import (
...
 
-			return true, errors.Wrapf(err, "Unable to get Tanzu Mission Control cluster entry, name : %s", d.Get(NameKey))
-		}
+type (
+	terraformConstruct string
+	operationType      string
+)
 
-		d.SetId(resp.Cluster.Meta.UID)
+const (
```
shreyassreenivas, line 40: Why do we need so many states, wouldn't it complicate things?
I feel that passing these caller type in the context would be a good way to pass to these functions.
and the values are 
"DataRead, RefreshState, CreateState, UpdateState", most cases there will be only "DataRead & RefreshState"

internal/resources/cluster/data_source_cluster.go
```diff
@@ -25,116 +25,162 @@ import (
 
 func DataSourceTMCCluster() *schema.Resource {
 	return &schema.Resource{
-		ReadContext: dataSourceTMCClusterRead,
+		ReadContext: schema.ReadContextFunc(dataSourceTMCClusterRead(withTerraformConstructType(constructTypeDataSource))),
```
shreyassreenivas, line 28: what is the advantage of this approach ?

internal/resources/cluster/data_source_cluster.go
```diff
@@ -25,116 +25,162 @@ import (
...
-		timeoutValueData = "3m"
+		timeoutValueData, _ := d.Get(waitKey).(string)
 
-		fallthrough
-	default:
-		timeoutDuration, parseErr := time.ParseDuration(timeoutValueData)
-		if parseErr != nil {
-			log.Printf("[INFO] unable to prase the duration value for the key %s. Defaulting to 3 minutes(3m)"+
-				" Please refer to 'https://pkg.go.dev/time#ParseDuration' for providing the right value", waitKey)
+		if cfg.operationType != operationCreate {
```
shreyassreenivas, line 111: all the context related logic should be written at one place for ease of reading /

internal/resources/cluster/data_source_cluster.go
```diff
@@ -25,116 +25,162 @@ import (
...
-		return diag.FromErr(err)
-	}
+		if err := d.Set(StatusKey, status); err != nil {
+			return diag.FromErr(err)
+		}
 
-	if err := d.Set(common.MetaKey, common.FlattenMeta(resp.Cluster.Meta)); err != nil {
-		return diag.FromErr(err)
-	}
+		if err := d.Set(common.MetaKey, common.FlattenMeta(resp.Cluster.Meta)); err != nil {
```
shreyassreenivas, line 172: check this indentation

internal/resources/cluster/data_source_cluster.go
ishangupta-ds: @shreyassreenivas we are using this for secrets as well (for some terraform construct type behaviours), it is already used for a different reason in policies, to keep things consistent. Also, I think functional parameters are always memory safe.

If you feel strongly that we should take the context-based approach do let us know for Kubernetes secret as well, since we have already referred to policy resource and made changes according to that.

shreyassreenivas: The source of invocation is not a functional parameter and something which can be referred from contexts we should leverage that. Contexts are meant to be used for these use cases.

I believe the present code compromises on the readability of the handlers and I would like to keep it plain and simple. 
I feel the source of the function call can be inferred from the context

Moreover, functional parameters should be used when we have multiple options being passed.

ishangupta-ds: Yeah, I agree, @Shubbhang351 let's take this MR as a reference for the k8's secret resource.

ishangupta-ds (approved):

internal/resources/cluster/data_source_cluster.go
```diff
@@ -45,6 +47,11 @@ func dataSourceTMCClusterRead(ctx context.Context, d *schema.ResourceData, m int
 		if err != nil {
 			if clienterrors.IsNotFoundError(err) {
 				_ = schema.RemoveFromState(d, m)
+
+				if ctx.Value(contextMethodKey{}) == "read" {
```
shreyassreenivas, line 51: you are erroring out after `_ = schema.RemoveFromState(d, m)`.. this is wrong
You should not remove it from the state and just error out in the case read context

internal/resources/cluster/data_source_cluster.go
```diff
@@ -25,12 +25,14 @@ import (
 
 func DataSourceTMCCluster() *schema.Resource {
 	return &schema.Resource{
-		ReadContext: dataSourceTMCClusterRead,
-		Schema:      clusterSchema,
+		ReadContext: func(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+			return dataSourceClusterRead(context.WithValue(ctx, contextMethodKey{}, "read"), d, m)
```
shreyassreenivas, line 29: lets be more explicit here and call it dataSrcRead

internal/resources/cluster/data_source_cluster.go
```diff
@@ -96,7 +103,7 @@ func dataSourceTMCClusterRead(ctx context.Context, d *schema.ResourceData, m int
 		_, err = helper.RetryUntilTimeout(getClusterResourceRetryableFn, 10*time.Second, timeoutDuration)
 	}
 
-	if err != nil || resp == nil {
+	if err != nil || resp == nil || resp.Cluster == nil {
```
shreyassreenivas, line 106: after fixing the above issues check is this extra validation is needed or not

internal/resources/cluster/data_source_cluster.go
Vasundharashukla: Done

internal/resources/cluster/data_source_cluster.go
Vasundharashukla: Done

internal/resources/cluster/data_source_cluster.go
Vasundharashukla: This is still needed, else the plugin crashes for the case of apply->remove from TMC (via diff client)->re-apply.

shreyassreenivas:
can you add an acceptance test for this scenario in the next MR

shreyassreenivas (approved):

Vasundharashukla:
> can you add an acceptance test for this scenario in the next MR

Sure

PR #140, opened 2023-03-15
Add read context for data source handler function for differentiating between different read calls

Vasundharashukla:
Differentiate between read calls of data source and resource read

        Fixes #

3. **Additional information**

Currently, PR supports changes for cluster group, node pool, credentials, workspace and namespace.
For, EKS and TSM integration, the changes shall be added after testing on connecting with @asutoshpalai and @johnweldon


internal/resources/cluster/nodepools/resource_node_pool.go
```diff
@@ -24,6 +24,8 @@ const (
 	ResourceName = "tanzu-mission-control_cluster_node_pool"
 )
 
+type contextMethodKey struct{}
```
shreyassreenivas, line 27: let's not duplicate the code and have the context creation in a lib and resued.
and in the same lib, have functions for creating a context with function name

internal/resources/credential/datasource_credential.go
```diff
@@ -39,7 +41,12 @@ func dataSourceCredentialRead(_ context.Context, d *schema.ResourceData, m inter
 	resp, err = config.TMCConnection.CredentialResourceService.CredentialResourceServiceGet(constructFullname(d))
 	if err != nil || resp == nil {
 		if clienterrors.IsNotFoundError(err) {
+			if ctx.Value(contextMethodKey{}) == "dataSourceRead" {
```
shreyassreenivas, line 44: create a constant for "dataSourceRead" and may be push all these logic in a lib

internal/resources/cluster/nodepools/resource_node_pool.go
Vasundharashukla: Please check the changes for nodepool resource, I will add the same to others once you approve these.

internal/helper/handler_context.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+const (
+	DataRead     Caller = "DataRead"
+	RefreshState Caller = "RefreshState"
+	CreateState  Caller = "CreateState"
+	UpdateState  Caller = "UpdateState"
+)
+
+func NewCallerContext(ctx context.Context) *callerContext {
```
shreyassreenivas, line 25: I do not think we need to create a struct 'callerContext' here.
All the helper functions are useful, but avoid the struct and the method binding

internal/resources/cluster/nodepools/data_source_node_pool.go
```diff
@@ -13,21 +13,40 @@ import (
...
+func dataSourceClusterNodePoolRead(ctx context.Context, d *schema.ResourceData, m interface{}) (diags diag.Diagnostics) {
 	config := m.(authctx.TanzuContext)
 
-	resp, err := config.TMCConnection.NodePoolResourceService.ManageV1alpha1ClusterNodePoolResourceServiceGet(constructFullName(d))
+	fn := constructFullName(d)
+	resp, err := config.TMCConnection.NodePoolResourceService.ManageV1alpha1ClusterNodePoolResourceServiceGet(fn)
+
+	if err != nil {
+		if clienterrors.IsNotFoundError(err) {
+			if helper.NewCallerContext(ctx).IsDataRead() {
```
shreyassreenivas, line 39: I think, we need to just flip the condition because we have a return with error below

Vasundharashukla: @shreyassreenivas I don't think we need to flip it based on similar check for cluster resource data read - [data_source_cluster.go#L55](https://github.com/vmware/terraform-provider-tanzu-mission-control/blob/143c1e710f639065d1924cfe57d479be64720867/internal/resources/cluster/data_source_cluster.go#L55)

internal/helper/handler_context.go
Vasundharashukla: Updated.

ishangupta-ds (approved):

internal/resources/cluster/nodepools/data_source_node_pool.go
```diff
@@ -13,21 +13,40 @@ import (
...
+func dataSourceClusterNodePoolRead(ctx context.Context, d *schema.ResourceData, m interface{}) (diags diag.Diagnostics) {
 	config := m.(authctx.TanzuContext)
 
-	resp, err := config.TMCConnection.NodePoolResourceService.ManageV1alpha1ClusterNodePoolResourceServiceGet(constructFullName(d))
+	fn := constructFullName(d)
+	resp, err := config.TMCConnection.NodePoolResourceService.ManageV1alpha1ClusterNodePoolResourceServiceGet(fn)
+
+	if err != nil {
+		if clienterrors.IsNotFoundError(err) {
+			if helper.IsDataRead(ctx) {
```
shreyassreenivas, line 39: here we need to flip the condition and the below return with error message will take care of this condition as well

internal/resources/cluster/nodepools/resource_node_pool.go
shreyassreenivas: Approved

internal/resources/cluster/nodepools/data_source_node_pool.go
shreyassreenivas: This error condition will be handled below, so flip the condition

ishangupta-ds (approved):

shreyassreenivas (approved):

PR #56, opened 2022-08-11
acceptance tests for security policy resource with minor fix

ishangupta-ds:
1. **What this PR does / why we need it**:  This PR [adds acceptance tests for security policy resource](https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/56/commits/70f0b630cc49b9d466ffec3356302111d9246fa6) with a minor fix: [add unknown type for scope and recipe in security policy resource](https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/56/commits/a986f7addb4ca9f20223171b44cf732bdea854ae)

2. **Additional information**: Tested for all cases.

3. **Screenshot of test results**:

Screenshot 2022-08-11 at 3 05 42 PM

PR #144, opened 2023-03-24
Add models, clients and schema for Git repository and Kustomization resource

ishangupta-ds:
**What this PR does / why we need it**:

Add models, clients and schema for Git repository and Kustomization resource


internal/resources/gitrepository/spec/spec.go
```diff
@@ -0,0 +1,153 @@
...
+	Description: "Spec for the Repository.",
+	Required:    true,
+	MaxItems:    1,
+	Elem: &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			URLKey: {
+				Type:         schema.TypeString,
+				Description:  "URL of the git repository. Repository URL should begin with http, https, or ssh",
+				Required:     true,
+				ValidateFunc: validation.IsURLWithScheme([]string{"http", "https", "ssh"}),
```
shreyassreenivas, line 29: I am not is this is a valid input with http .. check with the team

internal/resources/gitrepository/spec/spec.go
```diff
@@ -0,0 +1,153 @@
...
+				Type:        schema.TypeString,
+				Description: "Reference to the secret. Repository credential.",
+				Optional:    true,
+				Default:     "",
+			},
+			intervalKey: {
+				Type:        schema.TypeString,
+				Description: "Interval at which to check gitrepository for updates. This is the interval at which Tanzu Mission Control will attempt to reconcile changes in the repository to the cluster. A sync interval of 0 would result in no future syncs. If no value is entered, a default interval of 5 minutes will be applied as `5m`.",
+				Optional:    true,
+				Default:     "300s",
```
shreyassreenivas, line 41: can you mention it in minutes ?

internal/resources/gitrepository/spec/spec.go
```diff
@@ -0,0 +1,153 @@
...
+	flattenRefData[semverKey] = ref.Semver
+	flattenRefData[commitKey] = ref.Commit
+
+	return []interface{}{flattenRefData}
+}
+
+func HasSpecChanged(d *schema.ResourceData) bool {
+	updateRequired := false
+
+	switch {
```
shreyassreenivas, line 139: check on this logic ? 
You will be setting update required only if all the values have changed

internal/resources/kustomization/resource_kustomization.go
```diff
@@ -0,0 +1,51 @@
...
+		Description: "Name of Namespace.",
+		Required:    true,
+		ForceNew:    true,
+	},
+	orgIDKey: {
+		Type:        schema.TypeString,
+		Description: "ID of Organization.",
+		Optional:    true,
+		ForceNew:    true,
+		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
shreyassreenivas, line 43: explain the context where this will be the case ?

internal/resources/kustomization/scope/scope.go
```diff
@@ -0,0 +1,97 @@
...
+		return scopedFullnameData
+	}
+
+	data, _ := value.([]interface{})
+
+	if len(data) == 0 || data[0] == nil {
+		return scopedFullnameData
+	}
+
+	scopeData := data[0].(map[string]interface{})
```
shreyassreenivas, line 47: possible panic

internal/resources/kustomization/scope/scope.go
```diff
@@ -0,0 +1,97 @@
...
+
+	if v, ok := scopeData[commonscope.ClusterKey]; ok && slices.Contains(ScopesAllowed[:], commonscope.ClusterKey) {
```
shreyassreenivas, line 49: ScopesAllowed[:], this means the full slice right ? just specifying ScopesAllowed should be good

shreyassreenivas, line 49: can we avoid using v and v1.. it bring down the readability

internal/resources/kustomization/spec/cluster_scope.go
```diff
@@ -0,0 +1,73 @@
...
+		return spec
+	}
+
+	data, _ := value.([]interface{})
+
+	if len(data) == 0 || data[0] == nil {
+		return spec
+	}
+
+	specData := data[0].(map[string]interface{})
```
shreyassreenivas, line 27: panic

resource_templates/cluster_group_scoped_kustomization.tf
```diff
@@ -0,0 +1,34 @@
...
+    path = "/"
+
+    prune = true
+
+    interval = "5m" // can be "5s" etc.
+
+    target_namespace = "test"
+  }
+
+  status = {
```
shreyassreenivas, line 31: status cannot be set in the resource .. this should only be computed

internal/resources/gitrepository/resource_git_repository.go
```diff
@@ -0,0 +1,51 @@
...
+		Description: "Name of Namespace.",
+		Required:    true,
+		ForceNew:    true,
+	},
+	orgIDKey: {
+		Type:        schema.TypeString,
+		Description: "ID of Organization.",
+		Optional:    true,
+		ForceNew:    true,
+		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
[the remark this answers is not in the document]
ishangupta-ds: This is required because the org id is an optional parameter in the scope/fullname unlike the others, when the user does not provide its value, we get it from the backend in the response and the state file gets updated with the org id.

This causes a diff in each `apply`. So we have put this check to suppress the diff if the org id in the state file is the same as the one coming from the backend.

To add to the above, we cannot make this value as computed since this is something that the client sends and is configurable as per the API spec. It may not be good to mark this field as computed.

internal/resources/gitrepository/data_source_git_repository.go
```diff
@@ -0,0 +1,14 @@
...
+SPDX-License-Identifier: MPL-2.0
+*/
+
+package gitrepository
+
+import "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+func DataSourceGitRepository() *schema.Resource {
+	return &schema.Resource{
+		Schema: gitRepositorySchema,
```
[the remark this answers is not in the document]
ishangupta-ds: They have been added in https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/145/commits

internal/resources/gitrepository/spec/spec.go
ishangupta-ds: The API spec does allow `http` scheme and the same is also present in the description of the swagger-generated model which is directly picked up from the proto definition for this RPC.

internal/resources/gitrepository/scope/scope.go
```diff
@@ -0,0 +1,97 @@
...
+			flattenScopeData[commonscope.ClusterKey] = FlattenClusterGitRepositoryFullname(scopedFullname.FullnameCluster)
+		}
+	case commonscope.ClusterGroupScope:
+		if slices.Contains(ScopesAllowed[:], commonscope.ClusterGroupKey) {
+			name = scopedFullname.FullnameClusterGroup.Name
+			namespace = scopedFullname.FullnameClusterGroup.NamespaceName
+			orgID = scopedFullname.FullnameClusterGroup.OrgID
+			flattenScopeData[commonscope.ClusterGroupKey] = FlattenClusterGroupGitRepositoryFullname(scopedFullname.FullnameClusterGroup)
+		}
+	case commonscope.UnknownScope:
```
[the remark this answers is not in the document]
ishangupta-ds: @shreyassreenivas We have precedence set for this conditional please refer - [security policy's scope](https://github.com/vmware/terraform-provider-tanzu-mission-control/blob/143c1e710f639065d1924cfe57d479be64720867/internal/resources/policy/scope/scope_schema.go#L126), the same has been followed across all scope based resource implementations.

Adding a `default` may not add much value due to the fact that the scope is being parsed from the name of the key for the map in the schema. If it does not match any other predefined scopes then it gets set to `UnknownScope`.

Moreover, this way of handling the default/unknown case was recommended by the team earlier based on the IAM policy implementation - 

MR with the change which added this `Unknown` scope: https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/56#issuecomment-1211770600 

Commit: https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/56/commits/a986f7addb4ca9f20223171b44cf732bdea854ae

internal/resources/gitrepository/spec/spec.go
ishangupta-ds: If we mention this in minutes, the backend converts it into seconds and we have an unwanted diff.

Also, the UI & CLI sends the default value as `300s` for this field. 

Let me know if there are any concerns with mentioning it in seconds instead of minutes.

internal/resources/kustomization/resource_kustomization.go
ishangupta-ds: This is required because the org id is an optional parameter in the scope/fullname unlike the others, when the user does not provide its value, we get it from the backend in the response and the state file gets updated with the org id.

This causes a diff in each `apply`. So we have put this check to suppress the diff if the org id in the state file is the same as the one coming from the backend.

To add to the above, we cannot make this value as computed since this is something that the client sends and is configurable as per the API spec. It may not be good to mark this field as computed.

internal/resources/gitrepository/spec/spec.go
ishangupta-ds: I have verified this logic and manually tested all the edge cases after implementing the operation/context handlers. 

This routine returns `true` when one or more values change in the spec due to the `fallthrough`.

internal/resources/kustomization/scope/scope.go
ishangupta-ds: We have a length check for the data and a nil check for the first element right before this type assertion.

```
if len(data) == 0 || data[0] == nil {
	return scopedFullnameData
}
```

I have manually tested and verified the scope block multiple times. I did not encounter a panic yet.


Also, we have been following this in every other scope implementation -
[resources/policy/scope/scope_schema.go#L65](https://github.com/vmware/terraform-provider-tanzu-mission-control/blob/143c1e710f639065d1924cfe57d479be64720867/internal/resources/policy/scope/scope_schema.go#L65)

internal/resources/kustomization/spec/cluster_scope.go
ishangupta-ds: We have a length check for the data and a nil check for the first element right before this type assertion.

```
if len(data) == 0 || data[0] == nil {
	return spec
}
```

I have manually tested and verified the spec block multiple times. I did not encounter a panic yet.


Also, we have been following this in every other spec implementation -
[resources/policy/kind/security/spec_schema.go#L45](https://github.com/vmware/terraform-provider-tanzu-mission-control/blob/143c1e710f639065d1924cfe57d479be64720867/internal/resources/policy/kind/security/spec_schema.go#L45)

internal/resources/kustomization/scope/scope.go
ishangupta-ds: I think `slices.Contains` function expects a slice which does not have a specific size instead of an array with a fixed size. If I make the change it won't compile.

I got the below error on my IDE when I replaced `ScopesAllowed[:]` with `ScopesAllowed`.

> Cannot use 'ScopesAllowed' (type [2]string) as the type []string

resource_templates/cluster_group_scoped_kustomization.tf
ishangupta-ds: Agreed, updated the same.

internal/resources/kustomization/scope/scope.go
ishangupta-ds: Agreed, updated the same.

internal/resources/gitrepository/spec/spec.go
ishangupta-ds: As discussed, I have added a `DiffSuppressFunc` for the intervals which will allow users to specify the same in minutes as well.

The default value has been updated to `5m` as part of this change.

internal/resources/kustomization/resource_kustomization.go
ishangupta-ds: As discussed, I have removed the `org_id` attribute from the schema since the resources are scoped to `cluster` and `cluster_group`.

shreyassreenivas (approved):

Vasundharashukla (approved):

PR #165, opened 2023-06-01
Acceptance test for network policy resource

Vasundharashukla:
Acceptance test for network policy resource

2. **Screenshot of successful run:**

Screenshot 2023-06-01 at 12 43 40 PM

3. **Screenshot of successful run after adding update steps for each recipe:**

Screenshot 2023-06-07 at 4 47 43 PM

### DIFF
--- a/internal/resources/policy/kind/network/resource/resource_network_policy_test.go
+++ b/internal/resources/policy/kind/network/resource/resource_network_policy_test.go
@@ -1,0 +1,584 @@
+/*
+Copyright © 2023 VMware, Inc. All Rights Reserved.
+SPDX-License-Identifier: MPL-2.0
+*/
+
+package networkpolicyresource
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/pkg/errors"
+
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/authctx"
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/client/proxy"
+	policyorganizationmodel "github.com/vmware/terraform-provider-tanzu-mission-control/internal/models/policy/organization"
+	policyworkspacemodel "github.com/vmware/terraform-provider-tanzu-mission-control/internal/models/policy/workspace"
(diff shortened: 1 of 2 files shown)


ishangupta-ds (approved):

internal/resources/policy/kind/network/resource/resource_network_policy_test.go
```diff
@@ -0,0 +1,557 @@
...
+	}
+}
+
+func TestAcceptanceForNetworkPolicyResource(t *testing.T) {
+	testConfig := testGetDefaultAcceptanceConfig(t)
+
+	t.Log("start network policy resource acceptance tests!")
+
+	// Test case for network policy resource with allow-all recipe.
+	resource.Test(t, resource.TestCase{
```
ramya-bangera, line 64: In the acceptance test I see the workflow we are covering is create/delete(delete I believe is implicitly tested by terrform) of infra object. Since updateoperation is supported on this resource, can you evaluate if it's feasible to cover this in the acceptance test as it's a valid case? I agree we do cover it in manual testing, all I want to understand is can this be covered as part of the acc test?

shreyassreenivas:
What about the mock tests ?

internal/resources/policy/kind/network/resource/resource_network_policy_test.go
```diff
@@ -0,0 +1,557 @@
...
+	policykindNetwork "github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/policy/kind/network"
+	policyoperations "github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/policy/operations"
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/policy/scope"
+	testhelper "github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/testing"
+)
+
+const (
+	networkPolicyResource    = policykindNetwork.ResourceName
+	networkPolicyResourceVar = "test_network_policy"
+	networkPolicyNamePrefix  = "tf-ip-test"
```
shreyassreenivas, line 35: I belive this prefix should be changed to tf-np-test/ 
what does ip stands for ?

Vasundharashukla: Done

Vasundharashukla:
> What about the mock tests ?

Will be taking it up later as separate PR along with other enhancements for policies.

internal/resources/policy/kind/network/resource/resource_network_policy_test.go
Vasundharashukla: We are testing some update cases for the recipes as well now. Added extra testing steps and changing some values in the input spec for them, satisfies our case. Added the recent successful run screenshot on the PR description.

ramya-bangera: Looks good, thanks for addressing the same. We can also add more test steps in future like applying when there are no changes in config and apply goes through.

PR #166, opened 2023-06-01
Examples and resource templates for network policy resource

Vasundharashukla:

Examples and resource templates for network policy resource for all recipes and scopes.


ishangupta-ds (approved):

examples/resources/network_policy/resource_organization_custom-ingress_network_policy.tf
```diff
@@ -0,0 +1,128 @@
...
+      custom_ingress {
+        rules {
+          ports {
+            port     = "8443"
+            protocol = "TCP"
+          }
+          rule_spec {
+            custom_selector {
+              namespace_selector {
+                key   = "key1a"
```
shreyassreenivas, line 26: Since customer are using these examples and most of the time, they just copy paste it. 
so lets start providing example from real world scenario, ie copy paste of this terraform block should work in real scenario.

Vasundharashukla: These examples are exactly the same resource blocks that I used in the live working demo, hence these are tested and work perfectly.

shreyassreenivas: Key's value should be component right ? that is what was shown in demo

shreyassreenivas: The examples should have real values which will be used in production

Vasundharashukla: > Key's value should be component right ? that is what was shown in demo

That namespace selector block is different and common to all policies and present in all scripts after the spec for this recipe ends. Check lines: 55-69

Vasundharashukla: > The examples should have real values which will be used in production

These are generic values and can be anything, so added the value as generic as possible. This has validation on schema as a validRegex. Also, resource templates have placeholders that give better description of what needs to be used for each spec value.

shreyassreenivas: Instead of key1a and value1a, can you add the right namespace selector

Vasundharashukla: I believe these should be fine since it is just key and value not governed by any rules but the length and regex that has been validated on the schema level. Still, I have asked the policy team to give some example values. 
cc: @sbasu31

ramya-bangera (approved):
LGTM

PR #183, opened 2023-06-30
added support for EKS cluster import

asutoshpalai:

This PR adds importing an existing EKS cluster (in TMC) into Terraform's state for further management under Terraform.

        Fixes # OLYMP-38955

3. **Additional information**

Screenshot of successful test run
image

### DIFF
--- a/internal/client/ekscluster/cluster_resource.go
+++ b/internal/client/ekscluster/cluster_resource.go
@@ -6,8 +6,13 @@ SPDX-License-Identifier: MPL-2.0
 package ekscluster
 
 import (
+	"fmt"
+	"net/http"
 	"net/url"
 
+	"github.com/pkg/errors"
+
+	clienterrors "github.com/vmware/terraform-provider-tanzu-mission-control/internal/client/errors"
 	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/client/transport"
 	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/helper"
 	eksmodel "github.com/vmware/terraform-provider-tanzu-mission-control/internal/models/ekscluster"
@@ -39,6 +44,8 @@ type ClientService interface {
 	EksClusterResourceServiceDelete(fn *eksmodel.VmwareTanzuManageV1alpha1EksclusterFullName, force string) error
 
 	EksClusterResourceServiceGet(fn *eksmodel.VmwareTanzuManageV1alpha1EksclusterFullName) (*eksmodel.VmwareTanzuManageV1alpha1EksclusterGetEksClusterResponse, error)
+
+	EksClusterResourceServiceGetByID(id string) (*eksmodel.VmwareTanzuManageV1alpha1EksclusterGetEksClusterResponse, error)
 
 	EksClusterResourceServiceUpdate(request *eksmodel.VmwareTanzuManageV1alpha1EksclusterCreateUpdateEksClusterRequest) (*eksmodel.VmwareTanzuManageV1alpha1EksclusterCreateUpdateEksClusterResponse, error)
 }
@@ -92,6 +99,36 @@ func (c *Client) EksClusterResourceServiceGet(fn *eksmodel.VmwareTanzuManageV1al
 	err := c.Get(requestURL, clusterResponse)
 
 	return clusterResponse, err
+}
+
+/*
+EksClusterResourceServiceGetByID gets an eks cluster by its ID.
+*/
+func (c *Client) EksClusterResourceServiceGetByID(id string) (*eksmodel.VmwareTanzuManageV1alpha1EksclusterGetEksClusterResponse, error) {
+	queryParams := url.Values{
+		"query": []string{fmt.Sprintf("uid=\"%s\"", id)},
+	}
+
+	requestURL := helper.ConstructRequestURL(apiVersionAndGroup).AppendQueryParams(queryParams).String()
+	clusterListResponse := &eksmodel.VmwareTanzuManageV1alpha1EksclusterListEksClustersResponse{}
+
+	err := c.Get(requestURL, clusterListResponse)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(clusterListResponse.EksClusters) == 0 {
+		return nil, clienterrors.ErrorWithHTTPCode(http.StatusNotFound, errors.New("cluster list by ID was empty"))
+	}
+
+	if len(clusterListResponse.EksClusters) > 1 {
+		return nil, clienterrors.ErrorWithHTTPCode(http.StatusExpectationFailed, errors.New("cluster list by ID returned more than one cluster"))
+	}
+
+	clusterResponse := &eksmodel.VmwareTanzuManageV1alpha1EksclusterGetEksClusterResponse{}
+	clusterResponse.EksCluster = clusterListResponse.EksClusters[0]
+
+	return clusterResponse, nil
 }
 
 /*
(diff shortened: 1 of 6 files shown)


internal/resources/ekscluster/helpers.go
```diff
@@ -77,6 +77,18 @@ func structEqual[T any](a, b *T) bool {
 	return reflect.DeepEqual(a, b)
 }
 
+func mapEqual[K comparable, V any](a, b map[K]V) bool {
```
ramya-bangera, line 80: Can you add the comment to indicate purpose of adding this wrapper func?

internal/resources/ekscluster/resource_ekscluster.go
```diff
@@ -571,11 +581,49 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
...
+
+	if err = d.Set(RegionKey, resp.EksCluster.FullName.Region); err != nil {
+		return nil, errors.Wrapf(err, "Failed to set region for the cluster %s", resp.EksCluster.FullName.Name)
+	}
+
+	if err = d.Set(NameKey, resp.EksCluster.FullName.Name); err != nil {
+		return nil, errors.Wrapf(err, "Failed to set name for the cluster %s", resp.EksCluster.FullName.Name)
+	}
+
+	err = setResourceData(d, resp.EksCluster, npresp.Nodepools)
```
ramya-bangera, line 614: I am not sure about the behaviour here, just wanted to understand if we need to set ID in terraform state explicitly for imported resource ? I mean this  - d.SetId(id)

ramya-bangera:
Q: With this approach how are we planning to write the resource definition for import?

internal/client/ekscluster/cluster_resource.go
```diff
@@ -94,6 +101,36 @@ func (c *Client) EksClusterResourceServiceGet(fn *eksmodel.VmwareTanzuManageV1al
 	return clusterResponse, err
 }
 
+/*
+EksClusterResourceServiceGetByID gets an eks cluster by its ID.
+*/
+func (c *Client) EksClusterResourceServiceGetByID(id string) (*eksmodel.VmwareTanzuManageV1alpha1EksclusterGetEksClusterResponse, error) {
+	queryParams := url.Values{
+		"query": []string{fmt.Sprintf("uid=\"%s\"", id)},
```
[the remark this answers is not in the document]
asutoshpalai: For this API, `query` is just a string. But, `url.Values` is defined as `map[string][]string`, so this is a slice of strings here.

internal/client/ekscluster/cluster_resource.go
```diff
@@ -94,6 +101,36 @@ func (c *Client) EksClusterResourceServiceGet(fn *eksmodel.VmwareTanzuManageV1al
...
+	err := c.Get(requestURL, clusterListResponse)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(clusterListResponse.EksClusters) == 0 {
+		return nil, clienterrors.ErrorWithHTTPCode(http.StatusNotFound, errors.New("cluster list by ID was empty"))
+	}
+
+	if len(clusterListResponse.EksClusters) > 1 {
```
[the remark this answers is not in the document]
asutoshpalai: This is a sanity check since only one cluster should be returned as we filter by cluster ID.

internal/resources/ekscluster/resource_ekscluster.go
asutoshpalai: Since the ID that we are using for import matches the ID that we would set, we don't need to explicitly set the ID. Ref: https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/import

internal/client/ekscluster/cluster_resource.go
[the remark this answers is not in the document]
asutoshpalai: This is invoking a List with a cluster ID filter. Since the filter format is a bit tricky, I didn't want to expose just the List where other modules have to figure out the correct filter format.

asutoshpalai:
> Q: With this approach how are we planning to write the resource definition for import?

There are a few ways that can be explored here. Any path we take will have this approach at its core.
- Writing a script that generates it
- Terraform in its new releases added (experimental) support for `-generate-config-out` as a flag for the `import` command that can do that for us.

internal/resources/ekscluster/resource_ekscluster.go
```diff
@@ -85,6 +81,9 @@ var clusterSchema = map[string]*schema.Schema{
 		Description: "Wait timeout duration until cluster resource reaches READY state. Accepted timeout duration values like 5s, 45m, or 3h, higher than zero",
 		Default:     "default",
 		Optional:    true,
+		DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
ramya-bangera, line 84: This will be returning true always.. So do we need this condition still?

ramya-bangera:
LGTM

ramya-bangera (approved):

internal/resources/ekscluster/resource_ekscluster.go
ramya-bangera: I read it wrongly as the function returns false always.. I get the use case basically we want to suppress any diff for wait timeout.

internal/resources/ekscluster/resource_ekscluster.go
```diff
@@ -571,11 +581,49 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
...
+		return nil, errors.Wrapf(err, "Failed to set region for the cluster %s", resp.EksCluster.FullName.Name)
+	}
+
+	if err = d.Set(NameKey, resp.EksCluster.FullName.Name); err != nil {
+		return nil, errors.Wrapf(err, "Failed to set name for the cluster %s", resp.EksCluster.FullName.Name)
+	}
+
+	err = setResourceData(d, resp.EksCluster, npresp.Nodepools)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to set resourcee data during import for %s", resp.EksCluster.FullName.Name)
```
shobha2626, line 616: nit: resourcee -> resource

internal/resources/ekscluster/resource_ekscluster.go
```diff
@@ -571,11 +581,49 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
...
+
+	return []*schema.ResourceData{d}, nil
+}
+
 func handleClusterDiff(config authctx.TanzuContext, tmcCluster *eksmodel.VmwareTanzuManageV1alpha1EksclusterEksCluster, meta *objectmetamodel.VmwareTanzuCoreV1alpha1ObjectMeta, clusterSpec *eksmodel.VmwareTanzuManageV1alpha1EksclusterSpec) error {
 	updateCluster := false
 
 	if meta.Description != tmcCluster.Meta.Description ||
-		!reflect.DeepEqual(meta.Labels, tmcCluster.Meta.Labels) {
+		!mapEqual(meta.Labels, tmcCluster.Meta.Labels) {
```
shobha2626, line 626: nit question: why is this change needed?

asutoshpalai: When the labels field is empty in one and not set in the other, `reflect.DeepEqual` leads to a false positive of change.

ramya-bangera (approved):

PR #190, opened 2023-07-11
Fix the policy tests

ramya-bangera:

- New param is introduced for network policy mock testing, so fixed other policy tests to supply this additional parameters.
- Skip the failing custom-egress and custom-ingress network policy tests from mock testing

4. **Which issue(s) this PR fixes**

        Fixes #

5. **Additional information**

6. **Special notes for your reviewer**

### DIFF
--- a/internal/resources/policy/kind/network/resource/resource_network_policy_test.go
+++ b/internal/resources/policy/kind/network/resource/resource_network_policy_test.go
@@ -281,35 +281,36 @@ func TestAcceptanceForNetworkPolicyResource(t *testing.T) {
 
 	t.Log("Network policy resource acceptance test complete for deny-all-egress recipe!")
 
+	// TODO: Skip custom-egress and custom-ingress tests currently for mock until the test issue is fixed.
 	// Test case for network policy resource with custom-egress recipe.
 	resource.Test(t, resource.TestCase{
 		PreCheck:          testhelper.TestPreCheck(t),
 		ProviderFactories: testhelper.GetTestProviderFactories(testConfig.Provider),
 		CheckDestroy:      nil,
 		Steps: func() []resource.TestStep {
-			steps := []resource.TestStep{
-				{
-					Config: testConfig.getTestNetworkPolicyResourceBasicConfigValue(scope.WorkspaceScope, policykindNetwork.CustomEgressRecipe),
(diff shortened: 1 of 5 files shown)


internal/resources/policy/kind/network/resource/resource_network_policy_test.go
```diff
@@ -281,35 +281,36 @@ func TestAcceptanceForNetworkPolicyResource(t *testing.T) {
 
 	t.Log("Network policy resource acceptance test complete for deny-all-egress recipe!")
 
+	// TODO: Skip custom-egress and custom-ingress tests currently for mock unitl the issue: OLYMP-40111 is fixed.
```
ishangupta-ds, line 284: Can we omit the JIRA ref.

internal/resources/policy/kind/network/resource/resource_network_policy_test.go
```diff
@@ -281,35 +281,36 @@ func TestAcceptanceForNetworkPolicyResource(t *testing.T) {
...
-			}
-
 			if found {
-				steps = append(steps, resource.TestStep{
-					Config: testConfig.getTestNetworkPolicyResourceBasicConfigValue(scope.WorkspaceScope, policykindNetwork.CustomEgressRecipe, WithPodLabel("key3", "value3")),
-					Check:  testConfig.checkNetworkPolicyResourceAttributes(scope.WorkspaceScope, policykindNetwork.CustomEgressRecipe),
-				})
+				steps := []resource.TestStep{
+					{
+						Config: testConfig.getTestNetworkPolicyResourceBasicConfigValue(scope.WorkspaceScope, policykindNetwork.CustomEgressRecipe),
```
shobha2626, line 294: nit: There is function called skipTest. Is it possible to make use of it instead?

ramya-bangera: Skip is for individual tests , since here all tests had to be skipped I chose this route. This is temporary until we have fixed the tests so I think this should be good.

shobha2626 (approved):

internal/resources/policy/kind/network/resource/resource_network_policy_test.go
ramya-bangera: Removed the Jira ref

PR #201, opened 2023-07-20
Add Package repository resource (operation handlers), data source and high level schema with validations

Shubbhang351:

Add Package repository resource (operation handlers), data source and high level schema with validations

### DIFF
--- a/internal/helper/helper.go
+++ b/internal/helper/helper.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"log"
 	"os"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 const DEV = "DEV"
@@ -125,4 +127,15 @@ func BoolPointer(b bool) *bool {
 
 func StringPointer(s string) *string {
 	return &s
+}
+
+func UpdateDataSourceSchema(d *schema.Schema) *schema.Schema {
+	dv := &schema.Schema{
+		Type:        d.Type,
+		Description: d.Description,
+		Computed:    true,
+		Elem:        d.Elem,
+	}
+
+	return dv
 }
--- a/internal/resources/tanzupackagerepository/constant.go
+++ b/internal/resources/tanzupackagerepository/constant.go
@@ -12,6 +12,4 @@ const (
 	NamespaceKey   = "namespace_name"
 	disabledKey    = "disabled"
 	DataSourceRead = "dataSourceRead"
-
-	globalRepoNamespace = "tanzu-package-repo-global"
 )
--- a/internal/resources/tanzupackagerepository/constant.go
+++ b/internal/resources/tanzupackagerepository/constant.go
@@ -12,4 +12,6 @@ const (
 	NamespaceKey   = "namespace_name"
 	disabledKey    = "disabled"
 	DataSourceRead = "dataSourceRead"
+
+	globalRepoNamespace = "tanzu-package-repo-global"
 )
--- a/internal/resources/tanzupackagerepository/data_source_repository.go
+++ b/internal/resources/tanzupackagerepository/data_source_repository.go
@@ -1,0 +1,85 @@
+/*
+Copyright © 2023 VMware, Inc. All Rights Reserved.
+SPDX-License-Identifier: MPL-2.0
+*/
+
+package tanzupackagerepository
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/pkg/errors"
+
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/authctx"
+	clienterrors "github.com/vmware/terraform-provider-tanzu-mission-control/internal/client/errors"
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/common"
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/tanzupackagerepository/scope"
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/tanzupackagerepository/spec"
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/tanzupackagerepository/status"
+)
+
+func DataSourcePackageRepository() *schema.Resource {
+	return &schema.Resource{
+		Schema: getDataSourceSchema(),
+		ReadContext: func(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+			return dataPackageRepositoryRead(context.WithValue(ctx, contextMethodKey{}, DataSourceRead), d, m)
+		},
+	}
+}
+
+func dataPackageRepositoryRead(ctx context.Context, d *schema.ResourceData, m interface{}) (diags diag.Diagnostics) {
+	config := m.(authctx.TanzuContext)
+
+	packageRepositoryName, ok := d.Get(nameKey).(string)
+	if !ok {
+		return diag.Errorf("Unable to read package repository name")
+	}
+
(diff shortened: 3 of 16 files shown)


internal/resources/tanzupackagerepository/resource_package_repository.go
```diff
@@ -6,39 +6,331 @@ SPDX-License-Identifier: MPL-2.0
...
+		return diag.Errorf("Unable to read package repository name")
+	}
+
+	scopedFullnameData := scope.ConstructScope(d, packageRepositoryName, packageRepositoryNamespace)
+
+	if scopedFullnameData == nil {
+		return diag.Errorf("Unable to create Tanzu Mission Control package repository entry; Scope full name is empty")
+	}
+
+	var (
```
ramya-bangera, line 114: Is it required?

internal/resources/tanzupackagerepository/data_source_repository.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		return diag.FromErr(errors.Wrapf(err, "Unable to get Tanzu Mission Control package repository entry, name : %s", packageRepositoryName))
+	}
+
+	if err := d.Set(disabledKey, clusterScopeStatus.Disabled); err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(UID)
```
ramya-bangera, line 70: Do we need this here? It will be set as part of create right?

internal/resources/tanzupackagerepository/resource_package_repository.go
```diff
@@ -6,39 +6,331 @@ SPDX-License-Identifier: MPL-2.0
...
+			Spec:     repoSpec,
+		},
+	}
+
+	_, err = config.TMCConnection.ClusterPackageRepositoryService.RepositoryResourceServiceUpdate(pkgRepoReq)
+	if err != nil {
+		return diag.FromErr(errors.Wrapf(err, "Unable to update Tanzu Mission Control cluster package repository entry, name : %s", packageRepositoryName))
+	}
+
+	if d.HasChange(disabledKey) {
```
ramya-bangera, line 244: If the update request is only for disabling the repo, then will the code in L239 get executed? What will be the exact behaviour in that case?

internal/resources/tanzupackagerepository/data_source_repository.go
Shubbhang351: but if we need to get pkg repo but if we are not creating it from terraform then we gonna required this.

internal/resources/tanzupackagerepository/resource_package_repository.go
Shubbhang351: correct if the update request is only for disabling the repo then the update for spec and meta shouldn't be called
I've updated it appropriately and tested it accordingly.

Shubbhang351: updated PTLA

internal/resources/tanzupackagerepository/data_source_repository.go
ramya-bangera: Sorry I couldn't get this, do you mean the pkg repo that is not created via terraform?

Shubbhang351: yes

ramya-bangera: In order to get such resources into terraform state you have to use terraform import and not like this in my opinion

Shubbhang351: this common practice I've seen throughout the multiple resources

internal/helper/helper.go
```diff
@@ -126,3 +128,14 @@ func BoolPointer(b bool) *bool {
 func StringPointer(s string) *string {
 	return &s
 }
+
+func UpdateDataSourceSchema(d *schema.Schema) *schema.Schema {
+	dv := &schema.Schema{
```
ishangupta-ds, line 133: Can you check if this function already exists?

internal/resources/tanzupackagerepository/constant.go
```diff
@@ -12,4 +12,6 @@ const (
 	NamespaceKey   = "namespace_name"
 	disabledKey    = "disabled"
 	DataSourceRead = "dataSourceRead"
+
+	globalRepoNamespace = "tanzu-package-repo-global"
```
ishangupta-ds, line 16: Will it always be this?

internal/resources/tanzupackagerepository/resource_package_repository.go
```diff
@@ -6,39 +6,314 @@ SPDX-License-Identifier: MPL-2.0
...
+		return diag.Errorf("Unable to read package repository name")
+	}
+
+	scopedFullnameData := scope.ConstructScope(d, packageRepositoryName, packageRepositoryNamespace)
+
+	if scopedFullnameData == nil {
+		return diag.Errorf("Unable to create Tanzu Mission Control package repository entry; Scope full name is empty")
+	}
+
+	// nolint: dogsled
```
ishangupta-ds, line 202: Please fix the lint issue.

ishangupta-ds (changes requested):

internal/resources/tanzupackagerepository/constant.go
Shubbhang351: in some other namespace we can not create pkg repository.

internal/helper/helper.go
Shubbhang351: yes we do have, but in my fork, I don't have feat/pde branch so that's why I have to add it again
while merging I'll take care for this.

ishangupta-ds: Thanks, please create an issue to track this.

internal/resources/tanzupackagerepository/constant.go
ishangupta-ds: I mean, the global namespace for uTKG (TKGS on VC8) clusters might be different, please check how to get this global namespace from the backend APIs instead of hardcoding it.

ishangupta-ds:
Have we validated all scenarios that we had discussed to be fixed?

internal/resources/tanzupackagerepository/constant.go
Shubbhang351: Updated PTAL

internal/resources/tanzupackagerepository/resource_package_repository.go
```diff
@@ -6,39 +6,340 @@ SPDX-License-Identifier: MPL-2.0
...
+		return diag.Errorf("Unable to read package repository name")
+	}
+
+	scopedFullnameData := scope.ConstructScope(d, packageRepositoryName)
+
+	if scopedFullnameData == nil {
+		return diag.Errorf("Unable to create Tanzu Mission Control package repository entry; Scope full name is empty")
+	}
+
+	_, err := GetGlobalNamespace(config, scopedFullnameData, d)
```
ishangupta-ds, line 170: Use from state file.

internal/resources/tanzupackagerepository/resource_package_repository.go
```diff
@@ -6,39 +6,340 @@ SPDX-License-Identifier: MPL-2.0
...
```
ishangupta-ds, line 201: Use from the state file.

ishangupta-ds (changes requested):

internal/resources/tanzupackagerepository/resource_package_repository.go
Shubbhang351: updated

internal/resources/tanzupackagerepository/resource_package_repository.go
Shubbhang351: updated

ishangupta-ds (approved):

ramya-bangera (approved):

PR #223, opened 2023-08-08
Vmware/gavins/improve terraform schema validation

gshaw-pivotal:
Adds schema validation to enum keys for AKS clusters so that if a user provides a value for a key and that value is not specified in the associated enum, 'terraform plan' will exit with an error.

Adds schema validation to 'min_count' and 'max_count' keys of auto_scaling_config key of nodepool to ensure the number provided is positive (greater than or equal to 1).

Fixes #OLYMP-41160

3. **Additional information**
Was unable to find a way to unit test. Tested via executing 'terraform plan' on a tf file where invalid values were provided for each key that validation logic was added to. Screenshots showing the error message(s) output from this are included.

Testing enum keys contained within the nodepool:
Screenshot 2023-08-08 at 1 00 54 PM

Testing min_count and max_count keys within auto_scaling_config key within the nodepool:
Screenshot 2023-08-08 at 12 59 05 PM

Testing upgrade_channel key of auto_upgrade_config within cluster config:
Screenshot 2023-08-08 at 1 00 06 PM

Testing name and tier keys of sku within the cluster config:
Screenshot 2023-08-08 at 12 59 42 PM


internal/resources/akscluster/schema.go
```diff
@@ -545,12 +605,27 @@ var NodepoolSpecSchema = &schema.Schema{
 				Description: "The OS type of the nodepool",
 				Default:     aksmodel.VmwareTanzuManageV1alpha1AksclusterNodepoolOsTypeLINUX,
 				Optional:    true,
+				ValidateDiagFunc: func(v any, path cty.Path) diag.Diagnostics {
```
asutoshpalai, line 608: Nit: I would like this more if the `validateSelectedValue` method returned another method instead of accepting a format string.

Or I just don't like the format string, you can accept the field name and substitute it in a format string in that method. 

But I think returning a method is cleaner here if you want to have a generic validating function.

asutoshpalai (changes requested):

asutoshpalai (dismissed):

internal/resources/akscluster/schema.go
```diff
@@ -653,11 +728,40 @@ var AutoScaleConfig = &schema.Resource{
...
+	for _, name := range allowedValues {
+		if name == selectedValue {
+			return diags
+		}
+	}
+
+	return append(diags, diag.Errorf("%s must be one of: %v", selectedKey, allowedValues)[0])
+}
+
+func validateIntValueIsPositive(value int, errorMessage string) diag.Diagnostics {
```
ramya-bangera, line 758: Can you take a look if any of the existing validation methods from terraform-plugin-sdk can be used here: https://github.com/hashicorp/terraform-plugin-sdk/tree/main/helper/validation 

Atleast for int validation you can use it.

gshaw-pivotal: @ramya-bangera 

Took your request and rewrote the schema validation to use existing validators. Note that ValidateFunc and is deprecated so I had to wrap the built in validators in ToDiagFunc so as to be compatible with ValidateDiagFunc.

Updated the screenshots in the MR commentary to show the new error messages as the text has changed.

Please review and approve.

ramya-bangera (approved):

asutoshpalai (approved):

PR #234, opened 2023-08-16
Validate network config for aks cluster

gshaw-pivotal:
When network_plugin key is set to kubenet, then setting dns_service_ip and/or service_cidr keys is not supported / allowed and will now return an error during terraform apply.

There are several network settings the API always allows you to pass, however these fields are ignored if the network type is not azure. This results in a situation where the Terraform state does not match the provision infrastructure as some values that were passed were ultimately ignored.

Thus this MR adds validation logic to ensure that the 'dns_service_ip' and/or 'service_cidr' keys are only set when the 'network_plugin' key is set to 'azure'. If they are set when they should not be, then terraform apply fails with an error message.

3. **Additional information**

Unit tested and tested by running 'terraform apply' with both valid and invalid network configs.

Screenshot 2023-08-14 at 6 29 34 PM

Also updated the aks documentation to highlight this requirement.

### DIFF
--- a/docs/data-sources/akscluster.md
+++ b/docs/data-sources/akscluster.md
@@ -102,13 +102,13 @@ Required:
 
 Optional:
 
-- `dns_service_ip` (String) IP address assigned to the Kubernetes DNS service
+- `dns_service_ip` (String) IP address assigned to the Kubernetes DNS service. This key can only be set when the network_config.network_plugin key is set to 'azure'.
 - `docker_bridge_cidr` (String) A CIDR notation IP range assigned to the Docker bridge network
 - `load_balancer_sku` (String) Load balancer SKU
-- `network_plugin` (String) Network plugin
+- `network_plugin` (String) Network plugin. Set the value of this key to 'azure' if you want to specify values for network_config.dns_service_ip and/or network_config.service_cidr.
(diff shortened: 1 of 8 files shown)


asutoshpalai:
This is fine but won't this check fit better at schema level validation?

gshaw-pivotal:
That was our first approach, but the ValidateDiagFunc is limited to primitive data types and can only see the value of the key it is associated with. In this case the value of key 1 has an impact on key 2 and key 3 values. We could not find a way to do that using the schema validators / behaviors. So we applied the validation as early as possible in the 'apply' to check that it is a valid network config.

asutoshpalai:
Won't we be able to add a validation function to the parent of key 1, key 2 and key 3 where we have the full view?

gshaw-pivotal:
> Won't we be able to add a validation function to the parent of key 1, key 2 and key 3 where we have the full view?

If you are referring to using 'ValidateDiagFunc' on the schema then as per the documentation (https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#validatediagfunc) we can only use said function on primitives, which the parent key would not be.

asutoshpalai (approved):

ramya-bangera (approved):

PR #239, opened 2023-08-17
Suppress diff during terraform plan for aks when features are disabled

gshaw-pivotal:

For terraform spec where a feature config has a sub-key called 'enable' if that key's value is false, then that feature will be ignored during the apply. Thus we should suppress it during the plan step.

We don't want to see diffs on features that are disabled, so that terraform plans correctly describe the actions that will be take.

As an example given:
```
auto_scaling_config {
  enable    = false
  max_count = 3
  min_count = 1
}
```
any changes to max_count or min_count should be ignored since the feature is marked as disabled and updates to min and max count will have no effect. Thus there is no value in showing them in the diff generated by terraform's plan step.

3. **Additional information**

Tested via executing 'terraform plan' for an aks cluster.

Given a terraform file, where the enable key is set to true, the following is the diff from terraform plan:

Screenshot 2023-08-17 at 5 37 23 PM

Screenshot 2023-08-17 at 5 37 37 PM

With this change to the schema validation, when the enable keys are set to false the following diff is generated:

Screenshot 2023-08-18 at 12 39 33 PM

### DIFF
--- a/internal/resources/akscluster/schema.go
+++ b/internal/resources/akscluster/schema.go
@@ -394,18 +394,27 @@ var AddonConfig = &schema.Resource{
 			Description: "Keyvault secrets provider addon",
 			Optional:    true,
 			Elem:        AzureKeyvaulSecretsProviderConfig,
+			DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+				return suppressConfig(k, d)
+			},
 		},
 		monitorAddonConfigKey: {
 			Type:        schema.TypeList,
 			Description: "Monitor addon",
 			Optional:    true,
 			Elem:        MonitorAddonConfig,
+			DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+				return suppressConfig(k, d)
+			},
 		},
 		azurePolicyAddonConfigKey: {
 			Type:        schema.TypeList,
 			Description: "Azure policy addon",
 			Optional:    true,
 			Elem:        AzurePolicyAddonConfig,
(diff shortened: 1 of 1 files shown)


asutoshpalai:
Is this a general TF behaviour or something specific to AKS?

gshaw-pivotal:
@asutoshpalai 

> Is this a general TF behaviour or something specific to AKS?

To try and answer your question, the ultimate root cause is AKS, as for certain parts of the config, if the 'enabled' flag is set to false, then it (Azure / AKS) does not apply that to the cluster being created. Taking the 'auto_scaling_config' example, as the enabled key has a value of false, AKS ignores / does not apply the min and max.

However, when you execute a 'tf plan' terraform does not know this, thus when it informs the user of the changes that are going to be made it will include the parts of the config that are going to be ignored by AKS. Thus using the example, 'auto_scaling_config' will show up in the plan, implying that it will be applied to the cluster, but in the end is not as AKS says 'this is disabled, do not need to apply'. This causes a difference between what 'tf plan' says is going to happen and what actually happens during 'tf apply'.

It was requested by a user to find a way to remove these parts of the config from being shown in the output of 'tf plan' for Azure so as to make the proposed changes and actual changes line up better. It is hoped that doing so will give a better user experience. Hence this MR. We elected to use the 'DiffSupressFunc' on the Azure schema as it can find the 'enabled' keys, check their value and then instruct terraform to suppress that part of the tf file from the plan output.

I don't know if EKS / AWS has a similar situation where parts of the config within the tf file are ignored when 'tf apply' is executed because of an 'enabled' flag. This MR only focuses on the Azure / AKS side and its schema.

Hope this helps to answer your question.

asutoshpalai (approved):

ramya-bangera (approved):

PR #247, opened 2023-08-25
Add go build tag support to run specific acceptance test

shobha2626:
1. **What this PR does / why we need it**: Add go build tag support to run specific acceptance test

        Fixes #

3. **Additional information**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,10 @@ ifeq ($(TEST_FLAGS),)
 	TEST_FLAGS := -cover
 endif
 
+ifeq ($(BUILD_TAGS),)
+	BUILD_TAGS := 'akscluster cluster clustergroup credential ekscluster gitrepository iampolicy kustomization namespace custompolicy imagepolicy networkpolicy quotapolicy securitypolicy sourcesecret workspace'
+endif
+
 .PHONY: build clean-up test gofmt vet lint acc-test website-lint website-lint-fix
 
 default: build
@@ -54,7 +58,8 @@ lint: gofmt
 	golangci-lint run -c ./.golangci.yml ./internal/... .
 
 acc-test: export TF_ACC = true
-acc-test: test
+acc-test:
+	go test $(TEST_PKGS) -tags $(BUILD_TAGS)
 
 website-lint:
 	@echo "==> Checking website against linters..."
--- a/README.md
+++ b/README.md
@@ -126,6 +126,20 @@ Run the command:
 $ make acc-test
 ```
 
+To run the acceptance test specific to a resource make use of the build-tags.
+Build tag name is equivalent to the corresponding resource name.
+
+`
+Running acceptance test without explicitly setting `BUILD_TAGS` runs all the acceptance test by default.
+ To specifically run acceptances test of a resouces, set `BUILD_TAGS` value to correponding resource name.
+`
+
+For instance to run acceptance test for cluster-group and namespace resource
+```sh
+$ export BUILD_TAGS = "clustergroup namespace"
+$ make acc-test
+```
+
 ### Test provider changes locally
 Please make use of a unique path as provided in the `Makefile` while building the provider with changes 
 and kindly use the same path in the source while using the provider to test the local changes.
--- a/internal/resources/gitrepository/resource_git_repository_test.go
+++ b/internal/resources/gitrepository/resource_git_repository_test.go
@@ -1,3 +1,6 @@
+//go:build gitrepository
+// +build gitrepository
+
 /*
 Copyright © 2023 VMware, Inc. All Rights Reserved.
 SPDX-License-Identifier: MPL-2.0
@@ -6,17 +9,14 @@ SPDX-License-Identifier: MPL-2.0
 package gitrepository
 
 import (
-	"context"
 	"fmt"
 	"log"
 	"os"
 	"strings"
 	"testing"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 	"github.com/pkg/errors"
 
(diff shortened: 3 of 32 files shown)


ramya-bangera:
- Fix linters
- Fix test failure
- Update the makefile targets and the test.yml worklfow to include the gittags while running tests.
- Update the readme with the details of this change and how to run acceptance test with gittags.


@shobha2626  - Please always ensure to validate the automated checks are green before publishing it out for review with wider audience.

shobha2626:
> 
@ramya-bangera 

* Fixed linters
* Fixed test failures
* Updated the makefile targets but I don't think we should update test.yml worklfow to include the gittags while running tests that is because this test.yml runs on every pull request and push and we should run all the test here.
* Updated the readme with the details of this change and how to run acceptance test with gittags

ramya-bangera:
> @ramya-bangera
> 
> * Fixed linters
> * Fixed test failures
> * Updated the makefile targets but I don't think we should update test.yml worklfow to include the gittags while running tests that is because this test.yml runs on every pull request and push and we should run all the test here.
test.yml worklfow is for running unit tests and build tags are added to acceptance tests. So in my opinion we don't need to make any changes to this workflow. Am I missing something?
> * Updated the readme with the details of this change and how to run acceptance test with gittags

Makefile
```diff
@@ -37,6 +41,9 @@ test: gofmt terraform-fmt vet lint
 	go mod tidy
 	go test $(TEST_PKGS) $(TEST_FLAGS)
 
+test-tags:
```
ramya-bangera, line 44: For my understanding how exactly this will work?

README.md
```diff
@@ -126,6 +126,13 @@ Run the command:
 $ make acc-test
 ```
 
+To run the acc-test specific to a resource make use of the build-tags.
```
ramya-bangera, line 129: - You can mention that since we have the build tags in place, if anybody tries to run tests without specifying any tags then ideally no tests are run. So the only way to run them is by specifying the corresponding tags.
- For better understanding add a point that the build tags are nothing but the corresponding resource names

internal/resources/gitrepository/test_helper.go
```diff
@@ -0,0 +1,44 @@
+/*
+Copyright © 2023 VMware, Inc. All Rights Reserved.
+SPDX-License-Identifier: MPL-2.0
+*/
+
+package gitrepository
```
ramya-bangera, line 6: Why it is moved to separate file?

Makefile
shobha2626: For example, if I want to run acceptance test for cluster group alone. I need to set the required env variables and BUILD_TAG=clustergroup.

Then run "make test-tags"

internal/resources/gitrepository/test_helper.go
shobha2626: The functions/variables declared within files consisting of build tags are ignored by the build tool. Since these functions/variables are used at other places also we error saying undefined. 

To overcome this issue we need to move this commonly used functions/variables to a common place. Hence, the change.

README.md
shobha2626: Done

Makefile
ramya-bangera: Ok now I see BUILD_TAG is updated have all build tags. That's fine, initially it was empty and that's where the confusion raised. Thanks for taking care of it

README.md
ramya-bangera: - The second point is still missing, you can specify the build tags are nothing but corresponding resource names.

- Also since by default all tags are included in BUILD_TAGS, you can mention user can use the `make test-tags` target to run all acc tests by not setting BUILD_TAGS manually to selected resources.

README.md
```diff
@@ -126,6 +126,17 @@ Run the command:
 $ make acc-test
 ```
 
+To run the acceptance test specific to a resource make use of the build-tags.
+
+`Running acceptance test without specifying any tags doesn't run any acceptance test at all.
```
ramya-bangera, line 131: I see you changed  the acc-test target `acc-test: test-tags` so with that changing will this statement holds good?

README.md
shobha2626: Done

README.md
shobha2626: Made the change accordingly

internal/resources/gitrepository/resource_git_repository_test.go
```diff
@@ -29,25 +29,6 @@ import (
...
-	GitRepositoryResource       string
-	GitRepositoryResourceVar    string
-	GitRepositoryResourceName   string
-	GitRepositoryName           string
-	ScopeHelperResources        *commonscope.ScopeHelperResources
-	GitRepositoryDataSourceVar  string
-	GitRepositoryDataSourceName string
-	Namespace                   string
-}
-
```
ishangupta-ds, line 50: @shobha2626 why are we moving this out to the test helper?

ishangupta-ds (changes requested):

internal/resources/gitrepository/resource_git_repository_test.go
shobha2626: Reposting the comment - https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/247#discussion_r1311251798

> The functions/variables declared within files consisting of build tags are ignored by the build tool. Since these functions/variables are used at other places also we error saying undefined.
> 
> To overcome this issue we need to move this commonly used functions/variables to a common place. Hence, the change.

ishangupta-ds: @shobha2626 have we run all the acceptance tests after making these changes?

shobha2626: I have run the acceptances test for majority of the resources and made sure the tags work. Since, there isn't any test changes I don't see any other concern

ishangupta-ds: Please run it for source secret, git repository and kustomization resources and ensure they still pass. 

I see test code changes even though they are just changes to the file.

shobha2626: Mock test passes for these resources, they make use of the same variables/functions moved to the new file. Hence, I don't see the concern of the test failing due to the changes made in this PR

Makefile
```diff
@@ -21,6 +21,10 @@ ifeq ($(TEST_FLAGS),)
 	TEST_FLAGS := -cover
 endif
 
+ifeq ($(BUILD_TAGS),)
```
ramya-bangera, line 24: Is the comma added at the end is intentional?

ramya-bangera (commented):
LGTM

ishangupta-ds (approved):

internal/resources/gitrepository/resource_git_repository_test.go
ishangupta-ds: Thanks!

Makefile
shobha2626: Yes, it is to check if the env variable is not set

ramya-bangera (approved):

PR #250, opened 2023-08-29
Add wait timeout support to credential resource

ramya-bangera:
      This PR adds timeout field to credential resource with which use can wait for the specified time until the supported credential reaches VALID state

        Fixes #Issue reported over Jira

3. **Additional information**

- With default timeout of 3m
Screenshot 2023-08-29 at 5 10 53 PM

### DIFF
--- a/internal/resources/credential/constants.go
+++ b/internal/resources/credential/constants.go
@@ -6,6 +6,8 @@ SPDX-License-Identifier: MPL-2.0
 package credential
 
 const (
+	ResourceName = "tanzu-mission-control_credential"
+
 	NameKey              = "name"
 	specKey              = "spec"
 	statusKey            = "status"
@@ -21,5 +23,5 @@ const (
 	iamRoleARNKey    = "arn"
 	iamRoleExtIDKey  = "ext_id"
 	typeKey          = "type"
-	ResourceName     = "tanzu-mission-control_credential"
+	waitKey          = "ready_wait_timeout"
 )
--- a/internal/resources/credential/datasource_credential.go
+++ b/internal/resources/credential/datasource_credential.go
@@ -7,6 +7,9 @@ package credential
 
 import (
 	"context"
+	"log"
+	"strings"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -18,6 +21,8 @@ import (
 	credentialsmodels "github.com/vmware/terraform-provider-tanzu-mission-control/internal/models/credential"
 	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/common"
 )
+
+const defaultWaitTimeout = 3 * time.Minute
 
 func DataSourceCredential() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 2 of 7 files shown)


internal/resources/credential/resource_credential.go
```diff
@@ -43,6 +44,12 @@ var credentialSchema = map[string]*schema.Schema{
 		Computed:    true,
 		Elem:        &schema.Schema{Type: schema.TypeString},
 	},
+	waitKey: {
+		Type:        schema.TypeString,
+		Description: "Wait timeout duration until credential resource reaches VALID state. Accepted timeout duration values like 5s, 5m, or 1h, higher than zero.",
+		Default:     "default",
```
shobha2626, line 50: Default wait time is "default"? why can't we set a default time instead

internal/resources/credential/datasource_credential.go
```diff
@@ -39,14 +44,59 @@ func dataSourceCredentialRead(ctx context.Context, d *schema.ResourceData, m int
 
 	name, _ := d.Get(NameKey).(string)
 
-	resp, err = config.TMCConnection.CredentialResourceService.CredentialResourceServiceGet(constructFullname(d))
-	if err != nil || resp == nil {
-		if clienterrors.IsNotFoundError(err) && !helper.IsDataRead(ctx) {
-			_ = schema.RemoveFromState(d, m)
-			return diags
+	getCredentialResourceRetrybleFn := func() (retry bool, err error) {
```
shobha2626, line 47: nit: Typo getCredentialResourceRetrybleFn  => getCredentialResourceRetryableFunc

internal/resources/credential/datasource_credential.go
```diff
@@ -39,14 +44,59 @@ func dataSourceCredentialRead(ctx context.Context, d *schema.ResourceData, m int
...
+		resp, err = config.TMCConnection.CredentialResourceService.CredentialResourceServiceGet(constructFullname(d))
+		if err != nil || resp == nil {
+			if clienterrors.IsNotFoundError(err) && !helper.IsDataRead(ctx) {
+				_ = schema.RemoveFromState(d, m)
```
shobha2626, line 51: Can we add clear comment on what is the case we remove the state? Like why only when it's any operation other than read with not found error?

This would make the code more readable and understandable.

internal/resources/credential/datasource_credential.go
```diff
@@ -39,14 +44,59 @@ func dataSourceCredentialRead(ctx context.Context, d *schema.ResourceData, m int
...
+		timeoutData = helper.DoNotRetry
+	}
+
+	var timeoutDuration time.Duration
+
+	switch timeoutData {
+	case helper.DoNotRetry:
+		_, err = getCredentialResourceRetrybleFn()
+	case "default":
+		// nolint: ineffassign
```
shobha2626, line 81: Why is the linter disabled?

shobha2626 (changes requested):

internal/resources/credential/constants.go
```diff
@@ -6,6 +6,8 @@ SPDX-License-Identifier: MPL-2.0
 package credential
 
 const (
+	ResourceName = "tanzu-mission-control_credential"
```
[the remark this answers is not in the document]
ramya-bangera: No major changes here, I have just moved it to the top for better readability and grouped constants related to key's to one block

internal/resources/credential/datasource_credential.go
ramya-bangera: Fixed. Fyi-  I was just following the same naming conventioneer followed in existing resources with wait timeout

internal/resources/credential/datasource_credential.go
ramya-bangera: In case of data read even if the resource is not found we don't remove it from state, all other cases if resource is not found its removed from the state. This implementation remains the same across different resources in the repo and that's the reason same is followed here.I believe this is self explanatory, if we need to add comment then it has to be across all resources.

internal/resources/credential/datasource_credential.go
ramya-bangera: The reason is we assign some value to `timeoutDuration` in L82, which I actually overridden in L86/L92. I can in fact remove this case completely but I have just kept it for consistency with other resources. 

Assume we reach L82 then the value of `timeoutDuration` is overridden in L92 as L86 will hit parsing error due to `timeoutData = default`

More details about this linter can be found [here](https://pkg.go.dev/github.com/gordonklaus/ineffassign#section-readme)

internal/resources/credential/resource_credential.go
ramya-bangera: For having a consistent user experience as existing are using "default",

internal/resources/credential/datasource_credential.go
shobha2626: Yes it is explanatory, still good to have a comment at least for this resource.

internal/resources/credential/datasource_credential.go
shobha2626: Let's remove the code that is added for consistency reason. It is always good to have better and cleaner code than following consistency.

internal/resources/credential/resource_credential.go
shobha2626: Existing as in for cluster resource? 
I see in the switch case where we set a default time, cannot we set it here?

shobha2626 (changes requested):

internal/resources/credential/datasource_credential.go
ramya-bangera: Only specifying the default timeout as `default` is added here for consistency.  Which code are you asking to remove?

ramya-bangera: I have modified it to have the default timeout now.

shobha2626: Thank you for addressing

shobha2626 (approved):
LGTM

PR #264, opened 2023-09-23
Feat/tf schema converter

GilTeraSky:

This PR includes a custom generic converter that can convert TF Schema values to Swagger API Models and vice versa.

2. **Additional information**

With the suggested approach there is no need to flatten/construct TF Schema values and Swagger model values.
All the is needed is to define a mapping schema and create a matching converter for the Swagger API model we wish to convert to and from.

3. **Special notes for your reviewer**

In this PR I also made an example of how the converter would work by modifying the credentials resource accordingly.


internal/helper/converter/array_indexer.go
```diff
@@ -0,0 +1,34 @@
+package converter
```
ramya-bangera, line 1: nit: add copyright to all new files

internal/helper/converter/converter.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+const (
+	arrIndexExceededMsg = "array index exceeded"
+	arrIsNilMsg         = "array is nil"
+)
+
+// ### Public Funcs ###.
+
+func (converter *TFSchemaModelConverter[T]) ConvertTFSchemaToAPIModel(data *schema.ResourceData, buildOnlyKeys []string) (modelPtr T) {
```
ramya-bangera, line 27: Can you add a description to the public methods?

internal/helper/converter/help/swagger_model_hierarchy_example.go
```diff
@@ -0,0 +1,66 @@
+//go:build ignore
+// +build ignore
+
+package help
+
+/*
```
ramya-bangera, line 6: Is this an example usage? In that instead of adding the commented code can we include in a Readme in the converter directory?

GilTeraSky: Yeah I will put it in a readme after I finish with the data protection, I included it in the PR because I thought you might need it to understand the logics.

internal/helper/converter/converter.go
GilTeraSky: yes sure

internal/helper/converter/array_indexer.go
GilTeraSky: Done

internal/helper/converter/converter.go
GilTeraSky: Done

internal/helper/converter/help/swagger_model_hierarchy_example.go
ramya-bangera: Got it thanks

internal/helper/converter/help/swagger_model_hierarchy_example.go
```diff
@@ -0,0 +1,66 @@
...
+	S3ForcePathStyle bool `json:"s3ForcePathStyle,omitempty"`
+	S3URL string `json:"s3Url,omitempty"`
+}
+
+type ClusterObject struct {
+	ManagementClusterName string `json:"managementClusterName,omitempty"`
+	Name string `json:"name,omitempty"`
+	ProvisionerName string `json:"provisionerName,omitempty"`
+}
+
```
Shubbhang351, line 62: can you also add an example where we have a field of interface type, something like this
```
type AssignedGroupsObject struct {
	Cluster interface{} `json:"cluster,omitempty"
}

```

ramya-bangera (dismissed):

ramya-bangera (approved):

internal/helper/converter/help/swagger_model_hierarchy_example.go
Shubbhang351: we have covered this in recording so resolving this

PR #265, opened 2023-09-24
Feat/dataprotection-enable -- Initial client, models and tf schema

GilTeraSky:
       This PR includes the Swagger API Models & Clients along with the Terraform Schema Design for enabling data protection 
       on a cluster.

        -

3. **Additional information**
       -

       -

### DIFF
--- a/internal/client/cluster/dataprotection/dataprotection_resource.go
+++ b/internal/client/cluster/dataprotection/dataprotection_resource.go
@@ -1,0 +1,94 @@
+/*
+Copyright © 2023 VMware, Inc. All Rights Reserved.
+SPDX-License-Identifier: MPL-2.0
+*/
+
+package dataprotectionclient
+
+import (
+	"net/url"
+	"strconv"
+
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/client/transport"
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/helper"
+	dataprotectionmodels "github.com/vmware/terraform-provider-tanzu-mission-control/internal/models/cluster/dataprotection"
+)
+
+const (
+	apiVersionAndGroup = "v1alpha1/clusters"
+	dataProtectionPath = "dataprotection"
+)
+
+// New creates a new cluster resource service API client.
+func New(transport *transport.Client) ClientService {
+	return &Client{Client: transport}
+}
+
+/*
+Client for credentials resource service API.
+*/
+type Client struct {
+	*transport.Client
+}
+
+// ClientService is the interface for Client methods.
+type ClientService interface {
+	DataProtectionResourceServiceCreate(request *dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionCreateDataProtectionRequest) (*dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionCreateDataProtectionResponse, error)
+
+	DataProtectionResourceServiceDelete(fn *dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionFullName, destroyBackups bool) error
+
+	DataProtectionResourceServiceList(clusterName string) (*dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionListDataProtectionsResponse, error)
+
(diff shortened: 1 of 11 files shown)


internal/models/cluster/dataprotection/spec.go
```diff
@@ -0,0 +1,40 @@
...
+
+import "github.com/go-openapi/swag"
+
+// VmwareTanzuManageV1alpha1ClusterDataprotectionSpec The spec collects all the options for installing backup and restore solution into a Kubernetes cluster.
+//
+// swagger:model vmware.tanzu.manage.v1alpha1.cluster.dataprotection.Spec
+type VmwareTanzuManageV1alpha1ClusterDataprotectionSpec struct {
+
+	// List of Backup Locations to install in the cluster.
+	BackupLocationNames []string `json:"backupLocationNames"`
```
rnarenpujari, line 11: This field is technically not deprecated in the api but is not recommended to be used. Is it ok to drop this now or does it need to be deprecated in the api first?

rnarenpujari (commented):
LGTM

internal/models/cluster/dataprotection/spec.go
GilTeraSky: It is not a problem to get rid of it as long as we also don't let you update this through TF.
Currently it is configured part of the schema as a field to be filled.

ramya-bangera:
@GilTeraSky -
1.  The changes should go to a feature branch and not to main until the feature is completely ready.
2. Fix the linters and copyright from 2022-2023

Overall LGTM

internal/models/cluster/dataprotection/spec.go
rnarenpujari: Would prefer to have it removed since we don't want to encourage users to populate this. @eastaga wdyt?

internal/client/cluster/dataprotection/dataprotection_resource.go
```diff
@@ -0,0 +1,91 @@
...
+	*transport.Client
+}
+
+// ClientService is the interface for Client methods.
+type ClientService interface {
+	DataProtectionResourceServiceCreate(request *dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionCreateDataProtectionRequest) (*dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionCreateDataProtectionResponse, error)
+
+	DataProtectionResourceServiceDelete(fn *dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionFullName, destroyBackups bool) error
+
+	DataProtectionResourceServiceList(clusterName string) (*dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionListDataProtectionsResponse, error)
```
eastaga, line 39: Data Protection also support listing all clusters with data protection enabled via wildcard.
Are we not supporting similar in terraform?

internal/models/cluster/dataprotection/spec.go
eastaga: Yes it should be removed from TF.

internal/client/cluster/dataprotection/dataprotection_resource.go
GilTeraSky: Currently the data protection is based on an individual cluster an this is how I understood the scope of the task and therefore it is implemented this way.
Who can confirm the scope of the task?

internal/models/cluster/dataprotection/spec.go
GilTeraSky: Done

internal/client/cluster/dataprotection/dataprotection_resource.go
eastaga: @ramya-bangera @shobha2626 Are we supporting listing anywhere else in terraform?

shobha2626: As mentioned by @GilTeraSky, Data protection resource is based on individual cluster. That case, we don't need list implementation.

GilTeraSky: Actually I removed the datasource for enable data protection because initially we thought users will be using this data source to get the backup location which mentioned to be deperecated and requested to be removed from the resource.

rnarenpujari (approved):

PR #269, opened 2023-09-24
Feat/dataprotection-backup-schedule -- Initial TF resource & data source

GilTeraSky:
       This PR includes the Terraform resources for creating a backup schedule on a cluster.

        -

3. **Additional information**
       -

       -

### DIFF
--- a/internal/resources/cluster/backupschedule/resource_backup_schedule.go
+++ b/internal/resources/cluster/backupschedule/resource_backup_schedule.go
@@ -45,7 +45,7 @@ func ResourceBackupSchedule() *schema.Resource {
 func resourceBackupScheduleCreate(ctx context.Context, data *schema.ResourceData, m interface{}) (diags diag.Diagnostics) {
 	config := m.(authctx.TanzuContext)
 	model := tfModelResourceConverter.ConvertTFSchemaToAPIModel(data, []string{})
-	diags = validateSchema(model, BackupScope(data.Get(scopeKey).(string)))
+	diags = validateSchema(model, BackupScope(data.Get(ScopeKey).(string)))
 
 	if diags.HasError() {
 		return diags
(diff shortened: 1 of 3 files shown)


internal/resources/cluster/backupschedule/resource_backup_schedule.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil || getResponse.Schedule == nil {
+		return diag.Errorf("Couldn't read backup schedule. Name: %s, Cluster: %s", backupScheduleFn.Name, backupScheduleFn.ClusterName)
+	} else {
+		err = tfModelResourceConverter.FillTFSchema(getResponse.Schedule, data)
+
+		if err != nil {
+			return diag.Errorf("Couldn't read backup schedule. Name: %s, Cluster: %s", backupScheduleFn.Name, backupScheduleFn.ClusterName)
+		}
+
+		data.SetId(fmt.Sprintf("%s/%s", backupScheduleFn.ClusterName, backupScheduleFn.Name))
```
rnarenpujari, line 68: as mentioned in the data protection [PR](https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/270#pullrequestreview-1643567481) I don't think just cluster name is sufficient.

rnarenpujari (changes requested):

internal/resources/cluster/backupschedule/resource_backup_schedule.go
GilTeraSky: Let's decide what would be a valid UID for a cluster and I will change both to follow the same convention.
Pay attention that in this resource it is also concatenated with the Event name so if you have any suggestions to that also, it will be great

GilTeraSky: As commented in enable data protection, I decide to go with managment_cluster_name/provisioner_name/cluster_name/backup_schedule_name to keep it consistent with import command

rnarenpujari (approved):

PR #270, opened 2023-09-24
Feat/dataprotection-enable -- Initial TF resource & data source

GilTeraSky:
       This PR includes the Terraform resources for enabling data protection on a cluster.

        -

3. **Additional information**
       -

       -

### DIFF
--- a/internal/client/cluster/dataprotection/dataprotection_resource.go
+++ b/internal/client/cluster/dataprotection/dataprotection_resource.go
@@ -37,7 +37,7 @@ type ClientService interface {
 
 	DataProtectionResourceServiceDelete(fn *dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionFullName, destroyBackups bool) error
 
-	DataProtectionResourceServiceList(clusterName string) (*dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionListDataProtectionsResponse, error)
+	DataProtectionResourceServiceList(fn *dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionFullName) (*dataprotectionmodels.VmwareTanzuManageV1alpha1ClusterDataprotectionListDataProtectionsResponse, error)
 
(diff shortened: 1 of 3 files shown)


internal/resources/cluster/dataprotection/datasource_enable_data_protection.go
```diff
@@ -0,0 +1,89 @@
...
+			schemaBackupLocations = oldSpec[backupLocationNamesKey]
+		}
+
+		err = tfModelConverter.FillTFSchema(dataProtection, data)
+
+		if err != nil {
+			diags = diag.FromErr(err)
+		}
+
+		data.SetId(dataProtection.FullName.ClusterName)
```
rnarenpujari, line 81: tbh I don't understand this code well but I guess this is the unique identifier for the resource. Cluster name is insufficient for this since a cluster is identified uniquely by (org id, management cluster name, provisioner name, cluster name). Could alternatively use the cluster id or the data protection resource uid (should be under meta.uid)

rnarenpujari (changes requested):

internal/resources/cluster/dataprotection/datasource_enable_data_protection.go
GilTeraSky: Just tell me what would you prefer to be set as the resource UID in the TF state and I shall change it

rnarenpujari: @shobha2626 what is the convention for cluster scoped resources? I see 3 options:
- cluster full name + resource name
- cluster uid + resource name
- resource uid
cc @eastaga

rnarenpujari: @ramya-bangera do you have any recommendation for this?

shobha2626: resource ID should be a unique identifier, setting it to any such unique identifier should be good.

rnarenpujari (approved):

PR #273, opened 2023-09-26
Fix source secret data source tests

Shubbhang351:

Fix source secret data source tests

Output for Acceptance tests.
Screenshot 2023-09-26 at 10 47 34 AM

### DIFF
--- a/internal/resources/sourcesecret/data_sourcesecret_test.go
+++ b/internal/resources/sourcesecret/data_sourcesecret_test.go
@@ -42,39 +42,6 @@ func TestAcceptanceForSourceSecretDataSource(t *testing.T) {
 				Check:  testConfig.checkSourceSecretDataSourceAttributes(),
 			},
 			{
-				Config: testConfig.getTestSourceSecretDataSourceBasicConfigValue(commonscope.ClusterScope, spec.SSHKey, WithKnownhosts("somehostes")),
-				Check:  testConfig.checkSourceSecretDataSourceAttributes(),
-			},
-			{
-				Config: testConfig.getTestSourceSecretDataSourceBasicConfigValue(commonscope.ClusterGroupScope, spec.SSHKey),
-				Check:  testConfig.checkSourceSecretDataSourceAttributes(),
-			},
-			{
-				Config: testConfig.getTestSourceSecretDataSourceBasicConfigValue(commonscope.ClusterGroupScope, spec.SSHKey, WithKnownhosts("somehostes")),
-				Check:  testConfig.checkSourceSecretDataSourceAttributes(),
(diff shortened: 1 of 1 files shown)


internal/resources/sourcesecret/data_sourcesecret_test.go
```diff
@@ -41,39 +41,6 @@ func TestAcceptanceForSourceSecretDataSource(t *testing.T) {
 				Config: testConfig.getTestSourceSecretDataSourceBasicConfigValue(commonscope.ClusterScope, spec.SSHKey),
 				Check:  testConfig.checkSourceSecretDataSourceAttributes(),
 			},
-			{
```
ishangupta-ds, line 44: Don't we need this test? @Shubbhang351

internal/resources/sourcesecret/data_sourcesecret_test.go
```diff
@@ -41,39 +41,6 @@ func TestAcceptanceForSourceSecretDataSource(t *testing.T) {
...
-				Config: testConfig.getTestSourceSecretDataSourceBasicConfigValue(commonscope.ClusterGroupScope, spec.SSHKey, WithKnownhosts("somehostes")),
-				Check:  testConfig.checkSourceSecretDataSourceAttributes(),
-			},
-		},
-	},
-	)
-
-	t.Log("source secret data source acceptance test complete for SSH credential type")
-
-	// Test case for source secret data source with Username/Password credential type.
```
ishangupta-ds, line 62: Why are we removing the tests?

ishangupta-ds (changes requested):
Are the removed test cases not needed?

ramya-bangera:
@Shubbhang351 - Can you capture what was the issue and what is the fix here in description?

internal/resources/sourcesecret/data_sourcesecret_test.go
Shubbhang351: second tests I've added last time just to test spec SSH Type, which can be checked through update call so I've merged the tests for both spec type.

internal/resources/sourcesecret/data_sourcesecret_test.go
Shubbhang351: tests which I've removed they were covering one specific spec type which can be tests through update scenario that's why I've merged those tests in to one.

internal/resources/sourcesecret/data_sourcesecret_test.go
ishangupta-ds: Sure.

ramya-bangera (approved):

PR #276, opened 2023-09-27
Update cluster tests and avoid skipping all cluster tests

shobha2626:
1. **What this PR does / why we need it**:  Make changes to tests and don't skip all clusetr test

        Fixes #

3. **Additional information**

### DIFF
--- a/internal/resources/cluster/resource_cluster_test.go
+++ b/internal/resources/cluster/resource_cluster_test.go
@@ -53,36 +53,80 @@ func TestAcceptanceForAttachClusterResource(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					checkResourceAttributes(provider, clusterConfig["attachWithKubeConfig"]...),
 				),
+				SkipFunc: func() (bool, error) {
+					if os.Getenv("ATTACH_KUBECONFIG") == "" {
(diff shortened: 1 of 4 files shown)


internal/resources/cluster/resource_cluster_test.go
```diff
@@ -53,36 +53,79 @@ func TestAcceptanceForAttachClusterResource(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					checkResourceAttributes(provider, clusterConfig["attachWithKubeConfig"]...),
 				),
+				SkipFunc: func() (bool, error) {
```
ramya-bangera, line 56: Can you add a validation in the test and fail the test if required env variables are not set clearly indicating which env variable they are missing like it is done in EKS cluster or git repository resources?

ramya-bangera (changes requested):

internal/resources/cluster/resource_cluster_test.go
shobha2626: Failing the test, will exit completely. 
Instead added a detailed log message and skipping it, so that next cluster variance test can run

ankitsny (approved):

PR #278, opened 2023-09-28
Add fix for source secret resource

Shubbhang351:

Add fix for source secret resource

Problem :- Since for source secret Type(SSH / usernamePassword) server would return empty value for password / Identity so in a scenario where from TF spec does not change but some thing else changes let say Meta, then in PUT call we are forwarding the value of spec whatever we are getting from server directly which contain empty value for Password / Identity. which will cause issue from server side since empty value for Password / Identity is not allowed.


ishangupta-ds (approved):
@Shubbhang351 please validate all scenarios.

ramya-bangera (approved):

PR #280, opened 2023-10-03
Feat/dataprotection -- Tests & Docs

GilTeraSky:
1. **What this PR does / why we need it**

    This PR is the final in the series of Data Protection resources task.
    It includes tests and docs for data protection enablement, target location and backup schedule resources and data sources.

3. **Which issue(s) this PR fixes**

5. **Additional information**

6. **Special notes for your reviewer**
   
   For running the acceptance tests it is required to have 2 environment variables:
   TMC_MANAGED_CREDENTIALS_NAME - this variable should be a name of a TMC managed credentials which is required for 
   running the tests. Although it is possible to create a TMC managed credentials via the resource, there is a necessity for the 
   credentials to include a valid IAM role because when creating a target location with this type of credentials it fails if the IAM 
   role doesn't have permissions.

   AZURE_CREDENTIALS_NAME - this variable should be a name of a azure self provisioned credentials which is required for 
   running the tests. It is not supported creating azure self provisioned credentials for data protection use case.


vrabbi (approved):
LGTM

examples/resources/dataprotection/povider.tf
```diff
@@ -0,0 +1,16 @@
+terraform {
+  required_providers {
```
ramya-bangera, line 2: Is there a usage for this file?

ramya-bangera (approved):
@GilTeraSky - Add mock tests too for data protection feature, you can refer akscluster resource, but you can take it up after utkg work

GilTeraSky:
> @GilTeraSky - Add mock tests too for data protection feature, you can refer akscluster resource, but you can take it up after utkg work

We already agreed that with the Converter there is no need to mock anything and we should add tests for the converter in the future by me or someone else.

examples/resources/dataprotection/povider.tf
GilTeraSky: I put this in the folder so it will be easy to test from this folder, should I remove it?

ramya-bangera:
> We already agreed

When I say mock test, it Is mimicking the server with custom request/response so that we can cover even edge cases which we usually can't cover with the actual server. let me know if need more details

PR #292, opened 2023-10-11
Fix/credentials

GilTeraSky:
   This PR fixes the Credentials resource for Self Provisioned credentials.
   The problem occurs because the Credentials API doesn't return the values under the Data field and it caused the credentials 
   resource to be in a state where update is always needed (Because of missing values in the state)

3. **Additional information**
   In order to solve this issue, I implemented a bypass that keeps the old spec data and fill it back after the Converter is filling 
    the schema.
   This way, the status computed field is filled with values and the data field will maintain the values it received from the 
   Terraform config.

    Tested for both TMC Managed & Self Provisioned AWS

### DIFF
--- a/internal/resources/credential/datasource_credential.go
+++ b/internal/resources/credential/datasource_credential.go
@@ -96,15 +96,19 @@ func dataSourceCredentialRead(ctx context.Context, d *schema.ResourceData, m int
 
 	d.SetId(resp.Credential.Meta.UID)
 
-	if helper.IsDataRead(ctx) {
-		err = tfModelResourceConverter.FillTFSchema(resp.Credential, d)
+	specData := d.Get(specKey)
+	err = tfModelResourceConverter.FillTFSchema(resp.Credential, d)
 
-		if err != nil {
-			log.Println(err)
-
-			return diag.Errorf("Unable to get credentials. Schema conversion failed.")
-		}
+	if len(specData.([]interface{})) > 0 {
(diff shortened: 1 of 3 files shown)


internal/resources/credential/resource_credential.go
```diff
@@ -38,12 +38,6 @@ var credentialSchema = map[string]*schema.Schema{
 	},
 	common.MetaKey: common.Meta,
 	specKey:        credSpec,
-	statusKey: {
```
ramya-bangera, line 41: Can you paste the example plan in which case this field is showing updates? I understand this field is not really useful unless some automations rely on it to make some decision

GilTeraSky: I am resolving this as we already discussed it

ramya-bangera (approved):

vrabbi (approved):
looks great! 
LGTM

internal/resources/credential/datasource_credential.go
```diff
@@ -95,13 +95,20 @@ func dataSourceCredentialRead(ctx context.Context, d *schema.ResourceData, m int
...
+	if len(specData.([]interface{})) > 0 {
+		_ = d.Set(specKey, specData)
+	}
+
 	if err != nil {
 		log.Println(err)
 
 		return diag.Errorf("Unable to get credentials. Schema conversion failed.")
 	}
+	// }
```
shobha2626, line 111: nit: please remove this comment

PR #300, opened 2023-10-17
Feat/UTKG -- Client + Models + Schema + Converter Mapping

GilTeraSky:
    This PR includes the Client + Models + TF Schema + Converter Mapping for the UTKG resource.

2. **Which issue(s) this PR fixes**:

3. **Additional information**:

4. **Special notes for your reviewer**:
    As we agreed, overrides and cluster_variable fields will be TypeString and will json encoded string.


internal/resources/tanzukubernetescluster/schema.go
```diff
@@ -0,0 +1,402 @@
...
+	err := json.Unmarshal([]byte(value.(string)), &valueJSON)
+
+	if err != nil {
+		return diag.FromErr(errors.Wrapf(err, "Value is not a valid JSON string."))
+	}
+
+	return nil
+}
+
+func checkVariablesValues(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
Axpz, line 321: Can a single variable change be reflected by `terrform plan`? I think the answer is yes, right? For example?
`+ variable_a = "xxx"`

Axpz (approved):

PR #302, opened 2023-10-23
Feat/utkg tf resource

GilTeraSky:
     This PR includes the TF resource implementation for UTKG

3. **Additional information**
    In addition to the usual resource implementation, I also implemented validators for Cluster Class Variables and OpenAPI V3 
    Schema (I put it in the Helper folder for future use) which are being used to validate the Cluster Variables & Node Pools 
    Overrides (if provided).


ramya-bangera:
@GilTeraSky-  it will be great if you could resolve the merge conflicts

internal/resources/tanzukubernetescluster/resource_tanzu_kuberenetes_cluster.go
```diff
@@ -0,0 +1,424 @@
...
+
+	if err != nil {
+		return diag.FromErr(errors.Wrapf(err, "Couldn't create UTKG Cluster."))
+	}
+
+	request := &tanzukubernetesclustermodels.VmwareTanzuManageV1alpha1ManagementClusterProvisionerTanzukubernetesClusterData{
+		TanzuKubernetesCluster: model,
+	}
+
+	_, err = config.TMCConnection.TanzuKubernetesClusterResourceService.TanzuKubernetesClusterResourceServiceCreate(request)
```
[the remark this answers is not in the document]
ramya-bangera: FYI you can refer akscluster ekscluster  resource implementation

internal/client/tanzukubernetescluster/tanzukubernetescluster_resource.go
```diff
@@ -0,0 +1,93 @@
...
+
+/*
+Client for tanzu kubernetes cluster resource service API.
+*/
+type Client struct {
+	*transport.Client
+}
+
+// ClientService is the interface for Client methods.
+type ClientService interface {
```
ramya-bangera, line 37: Add client for nodepool service also

internal/resources/tanzukubernetescluster/resource_tanzu_kuberenetes_cluster.go
```diff
@@ -0,0 +1,424 @@
...
+	return err
+}
+
+// Helper functions.
+
+func readResourceWait(ctx context.Context, config *authctx.TanzuContext, resourceFullName *tanzukubernetesclustermodels.VmwareTanzuManageV1alpha1ManagementClusterProvisionerTanzukubernetesClusterFullName) (resp *tanzukubernetesclustermodels.VmwareTanzuManageV1alpha1ManagementClusterProvisionerTanzukubernetesClusterData, err error) {
+	stopStatuses := map[tanzukubernetesclustermodels.VmwareTanzuManageV1alpha1CommonClusterStatusPhase]bool{
+		tanzukubernetesclustermodels.VmwareTanzuManageV1alpha1CommonClusterStatusPhaseERROR:         true,
+		tanzukubernetesclustermodels.VmwareTanzuManageV1alpha1CommonClusterStatusPhaseUPGRADEFAILED: true,
+		tanzukubernetesclustermodels.VmwareTanzuManageV1alpha1CommonClusterStatusPhaseREADY:         true,
```
ramya-bangera, line 276: @GilTeraSky - We discussed about this on Friday and today we had an internal discussion so based on that here are the changes we would request you to do.

Schema should include the following
1. By default as per your implementation the operations would wait on these stop statuses. A minor change I would request you is to look for status phase `VmwareTanzuManageV1alpha1CommonClusterStatusPhaseHEALTHY` rather than `VmwareTanzuManageV1alpha1CommonClusterStatusPhaseREADY`. But in some unexpected cases if at all this can lead to infinite wait loop, lets ensure that we have  wait_timeout field in the schema with a  default timeout (if not specified by the user with which we can terminate the wait loop in worst scenarios, like the one you observed with TKGm delete in acceptance tests).
2. Add 2 more fields to the schema a bool `waitForKubeConfig` and `kubeconfig`(computed field which is for printing the kubeconfig as output). The second field would be set only when the `waitForKubeConfig` is true.
For reference you can see this  [PR](https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/313/). This one has been an ask from our terraform users , so lets try to address this

cc: @vmw-vjn

internal/resources/tanzukubernetescluster/resource_tanzu_kuberenetes_cluster.go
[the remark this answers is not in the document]
GilTeraSky: but the cluster API handles nodepools and as you mentioned previously on the design doc, we are going with a single resource for both cluster and nodepool and not separating them. why do we need to add the complexity of an additional client when the Cluster API provides all needed capabilities and no additional resource is being created to manage nodepools separately? this seems like over kill for this task unless i am missing something.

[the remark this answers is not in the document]
Axpz: Yes Ramya,  

> FYI you can refer akscluster ekscluster resource implementation

internal/resources/tanzukubernetescluster/resource_tanzu_kuberenetes_cluster.go
vmw-vjn: Thanks @ramya-bangera . Only thing I'll add (may be obvious, so sorry if its repetitive) is that:
said wait_timeout field value (whether it be default or user-provided), should be the main counter to terminate all/any wait loops - including for `VmwareTanzuManageV1alpha1CommonClusterStatusPhaseHEALTHY ` as well as `waitForKubeConfig `

internal/resources/tanzukubernetescluster/resource_tanzu_kuberenetes_cluster.go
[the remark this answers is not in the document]
ramya-bangera: > but the cluster API handles nodepools and as you mentioned previously on the design doc, we are going with a single resource for both cluster and nodepool and not separating them. why do we need to add the complexity of an additional client when the Cluster API provides all needed capabilities and no additional resource is being created to manage nodepools separately? this seems like over kill for this task unless i am missing something.

Today it looks like overkill because you see that Cluster API is able to handle nodepool operations but it will not work in the long run because the LCM team roadmap is to stop supporting nodepool operations via cluster API and only allow nodepool operations via nodepool APis . Thats why you see those 2 options of whether to have 2 resources(cluster and nodepool)/single resource(cluster including nodepool) is documented.

internal/resources/tanzukubernetescluster/resource_tanzu_kuberenetes_cluster.go
GilTeraSky: Just to clarify - if the timeout is expired, should I fail the resource?

vmw-vjn: If timeout is expired, then I'd say the should exit based on resource (cluster) state.

- Cluster is in healthy state at timeout but pinniped is not healthy. Exit with resource creation success but kubeconfig output set as null; assuming that `waitforKubeConfig` flag has been set by user. Optionally a warning or message stating that timeout occurred and KubeConfig couldn't be retrieved in set time (for example).
- Cluster is in ready state, not yet healthy at timeout. Exit with resource creation success, but warning or message stating resource was created but may not be healthy (for example).
- Cluster is still processing state, not yet ready at timeout. Exit with resource creation fail and rollback/clean-up where possible. Warning or message stating resource could not be successfully created in the set time (for example)

@ramya-bangera - please comment if I've got something incorrect or missed a scenario in list

vmw-vjn: As agreed today in Zoom call, I believe the path forward is to implement another flag for setting timeout behavior. If set to true (default), on time-out the resource creation/change  loop is exited and resource creation is failed in TF and if set to false, on time-out the resource creations loop is exited and resource creation is not failed in TF, if cluster is still in processing or ready or deploying stages.

internal/resources/tanzukubernetescluster/helper.go
```diff
@@ -0,0 +1,495 @@
...
+// This function determines whether a timeout is needed and whether to fail the request if a deadline has exceeded.
+func readResourceWait(ctx context.Context, config *authctx.TanzuContext, clusterFn *tanzukubernetesclustermodels.VmwareTanzuManageV1alpha1ManagementClusterProvisionerTanzukubernetesClusterFullName, existingNodePools []*tkcnodepoolmodels.VmwareTanzuManageV1alpha1ManagementClusterProvisionerTanzukubernetesClusterNodepool, timeoutPolicy map[string]interface{}) (resp *tanzukubernetesclustermodels.VmwareTanzuManageV1alpha1ManagementClusterProvisionerTanzukubernetesClusterData, err error) {
+	timeout := timeoutPolicy[TimeoutKey].(int)
+	waitForKubeConfig := timeoutPolicy[WaitForKubeConfigKey].(bool)
+	failOnTimeOut := timeoutPolicy[FailOnTimeOutKey].(bool)
+
+	if timeout > 0 {
+		var cancel context.CancelFunc
+
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(timeout)*time.Minute)
```
Axpz, line 40: Does this timeout has conflict with original timeout defined by client?

GilTeraSky: No because I disabled the client's timeout.
See the resource object, all operations are declared without timeout.

vrabbi (approved):
LGTM
Amazing work!

Axpz (approved):

PR #304, opened 2023-10-24
Secondary IP for AKS

grenader:
Add support for the 'Secondary IP' feature.

We have added:
- two new fields: PodSubnetID and NetworkPluginMode 
- various validations to make sure the correct information was passed to MKP
- unit tests to verify extended valuation 
- updated the documentation
- corrected schema for data source

The validations are:
- Network-plugin-mode 'overlay' can be only set when network-plugin is 'azure' - DONE 
- If network-plugin-mode 'overlay' is selected pod-subnet-id should NOT be defined in node pools.
- Node (Vnet-subnet) and pod subNets should belong to the same vNet
- pod subNet cannot be specified if node subNet is not defined
- node (Vnet-subnet) and pod subNets should be different from each other, this is in every node pool and across the whole cluster
- podCIDR can be only set for network-plugin is 'kubenet' or network-plugin is 'azure' with network-plugin-mode is 'overlay

(https://jira.eng.vmware.com/browse/TMC-43293)

3. **Additional information**

In addition to the 'secondary IP' field validation, the 
"DNS Server Ip or service CIDR cannot be set is network plugin is 'kubenet'" restriction has been removed, to follow Azure's actual behavior.


gshaw-pivotal (approved):

internal/models/akscluster/cluster_s_k_u_name.go
```diff
@@ -33,6 +33,9 @@ const (
 
 	// VmwareTanzuManageV1alpha1AksclusterClusterSKUNameBASIC captures enum value "BASIC".
 	VmwareTanzuManageV1alpha1AksclusterClusterSKUNameBASIC VmwareTanzuManageV1alpha1AksclusterClusterSKUName = "BASIC"
+
+	// VmwareTanzuManageV1alpha1AksclusterClusterNetworkPluginModeOverlay captures value "overlay".
+	VmwareTanzuManageV1alpha1AksclusterClusterNetworkPluginModeOverlay VmwareTanzuManageV1alpha1AksclusterClusterSKUName = "overlay"
```
TaylorOno, line 38: can we delete this from here

TaylorOno (approved):

asutoshpalai (approved):
LGTM

PR #310, opened 2023-10-30
Feat/ClusterClass -- Data Source

GilTeraSky:
    This PR includes a data source only for cluster class.
3. **Additional information**
    This data source wasn't initially planned but after some discussions with Scott we decided it will be a good idea to create a 
    data source which will simplify the user experience with the UTKG resource.
    The data source is useful for a user to understand what are the allowed worker classes, how the OpenAPIV3 schema looks 
    like and to get an example of how to build the Cluster Variables by returning an example template which essentially built 
    according to the OpenAPIV3 schema.

    This PR was taken from the UTKG resources branch and should be merged after the UTKG resources is merged.


vrabbi (approved):
LGTM! this looks great, thanks for the hard work on this. the UX around this is very good, and offers a great path forwards when custom cluster classes are supported in TMC

Axpz:
It's great, thanks.

PR #314, opened 2023-10-30
Feat/Custom IAM Role -- TF Resources

GilTeraSky:
   This PR includes the Terraform Resource implementation for the Custom IAM role.

3. **Additional information**
   a. In the read function I implemented a fix for API inconsistencies which include:
      a.1 MatchExpression (AggregationRule) doesn't return from the API, therefore I implemented a merge between the 
            response and the tf data to avoid changes but keeping new data from response (if exists).
      a.2 Sometimes the order of the array returning for the Resources is not the same but containing the same data, therefore I 
           implemented a merge between the response and the tf data to avoid changes but keeping new data from response (if 
           exists).
    b. I implemented a schema validation during the plan phase to make sure resources or url_paths (Exactly one) exist in each of 
        the Rules provided.

### DIFF
--- a/internal/client/customiamrole/customiamrole_resource.go
+++ b/internal/client/customiamrole/customiamrole_resource.go
@@ -21,7 +21,7 @@ func New(transport *transport.Client) ClientService {
 }
 
 /*
-Client for credentials resource service API.
+Client for custom iam role resource service API.
 */
 type Client struct {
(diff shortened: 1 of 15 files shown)


iprathapan (approved):

vrabbi (approved):
LGTM

Issue #181, opened 2023-06-27
EKS Cluster Ready Wait Timeout should default to 30m, but is actually 3m


doddatpivotal:
### Describe the bug

Version 1.1.7

Documentation provided at the terraform marketplace says that the `ready_wait_timeout` has a default value of 30m, however that is not what is happening.  It is much shorter.

I have found at https://github.com/vmware/terraform-provider-tanzu-mission-control/blob/main/internal/resources/ekscluster/resource_ekscluster.go#L34 that the code sets the default to 3m.

### Reproduction steps

1. Create an eks cluster.  After about 3 minutes, it will report back successful.
2. Go to TMC UI and see the cluster is still creating.

### Expected behavior

The terraform provider should wait up to 30m for the eks cluster to be ready before reporting back status.

### Additional context

https://github.com/vmware/terraform-provider-tanzu-mission-control/blob/main/resource_templates/ekscluster.tf has indicated in comments that 30m is the default.

3m is not realistic to expect a EKS cluster to be created.  In my experience a small cluster with 3 nodes takes about 12 minutes.

vmw-vjn:
Assigning to @asutoshpalai - to continue the conversation/investigation.

PR #324, opened 2023-11-02
Set default wait to 30 minutes for eks cluster create

gshaw-pivotal:

Standardizes the ready wait timeout for EKS cluster creation to 30 minutes.

Introduces 2 constants:
- a constant to hold the timeout in minutes (30)
- a constant to hold the timeout in nano seconds (the base time unit) and derives its value from the first constant
To change the default timeout one only needs to change the value in the 'minutesBasedDefaultTimeout' constant.

Fixes logic in 'internal/resources/ekscluster/data_source.go' where if a user did not provide a 'ready_wait_timeout' key, or provided the value of 'default' it now sets the timeout to 30 minutes instead of 5 minutes.

Revisits Issue #181 

3. **Additional information**

### DIFF
--- a/internal/resources/ekscluster/data_source.go
+++ b/internal/resources/ekscluster/data_source.go
@@ -8,6 +8,7 @@ package ekscluster
 import (
 	"context"
 	"log"
+	"strconv"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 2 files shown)


shobha2626 (approved):
LGTM

asutoshpalai (approved):

PR #328, opened 2023-11-07
add ability to update default node pool for tkg

warroyo:

This adds the logic needed to be able to update certain settings on the default node pool through the cluster resource for TKGs and TKGm. This is not for cluster class based clusters.

### DIFF
--- a/internal/resources/cluster/resource_cluster.go
+++ b/internal/resources/cluster/resource_cluster.go
@@ -589,7 +589,6 @@ func withMetaUpdate(d *schema.ResourceData, cluster *clustermodel.VmwareTanzuMan
 }
 
 func withTKGNodePoolUpdate(d *schema.ResourceData, nodepool *nodepoolmodel.VmwareTanzuManageV1alpha1ClusterNodepoolNodepool) bool {
-
 	switch {
 	case nodepool.Spec.TkgServiceVsphere != nil:
 		nodepools := d.Get(helper.GetFirstElementOf(SpecKey, tkgServiceVsphereKey, topologyKey, nodePoolKey)).([]interface{})[0].(map[string]interface{})
@@ -615,6 +614,7 @@ func withTKGNodePoolUpdate(d *schema.ResourceData, nodepool *nodepoolmodel.Vmwar
 			if incomingTkgServiceVsphereStorageClass != "" {
 				nodepool.Spec.TkgServiceVsphere.StorageClass = incomingTkgServiceVsphereStorageClass
 			}
+
 			log.Printf("[INFO] updating TKGs workload cluster nodepools")
 			return true
 		}
@@ -664,7 +664,7 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
 
 		log.Printf("[INFO] cluster update successful")
 	}
-	//check default nodepool configuration update
+	// check default nodepool configuration update
 	npFullName := constructNpFullName(d)
 	npResp, err := config.TMCConnection.NodePoolResourceService.ManageV1alpha1ClusterNodePoolResourceServiceGet(npFullName)
 	if err != nil {
--- a/internal/resources/cluster/resource_cluster.go
+++ b/internal/resources/cluster/resource_cluster.go
@@ -595,6 +595,7 @@ func withTKGNodePoolUpdate(d *schema.ResourceData, nodepool *nodepoolmodel.Vmwar
 		poolSpec := nodepools[SpecKey].([]interface{})[0].(map[string]interface{})
 		// poolInfo := nodepools["info"].([]interface{})[0].(map[string]interface{})
 		tkgsSpec := poolSpec[tkgServiceVsphereKey].([]interface{})[0].(map[string]interface{})
+
 		if d.HasChange(helper.GetFirstElementOf(SpecKey, tkgServiceVsphereKey, topologyKey, nodePoolKey)) {
 
 			incomingWorkerNodeCount := poolSpec[workerNodeCountKey].(string)
@@ -616,17 +617,21 @@ func withTKGNodePoolUpdate(d *schema.ResourceData, nodepool *nodepoolmodel.Vmwar
 			}
 
 			log.Printf("[INFO] updating TKGs workload cluster nodepools")
+
 			return true
 		}
 	case nodepool.Spec.TkgVsphere != nil:
 		nodepools := d.Get(helper.GetFirstElementOf(SpecKey, tkgVsphereClusterKey, topologyKey, nodePoolKey)).([]interface{})[0].(map[string]interface{})
 		poolSpec := nodepools[SpecKey].([]interface{})[0].(map[string]interface{})
+
 		if d.HasChange(helper.GetFirstElementOf(SpecKey, tkgVsphereClusterKey, topologyKey, nodePoolKey)) {
 			incomingWorkerNodeCount := poolSpec[workerNodeCountKey].(string)
 
 			if incomingWorkerNodeCount != "" {
 				nodepool.Spec.WorkerNodeCount = incomingWorkerNodeCount
 			}
+
+			return true
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


internal/resources/cluster/resource_cluster.go
```diff
@@ -100,6 +102,29 @@ func constructFullname(d *schema.ResourceData) (fullname *clustermodel.VmwareTan
...
+
+	if value, ok := d.GetOk(ProvisionerNameKey); ok {
+		fullname.ProvisionerName = value.(string)
+	}
+
+	if value, ok := d.GetOk(NameKey); ok {
+		fullname.ClusterName = value.(string)
+	}
+
+	if value, ok := d.Get(helper.GetFirstElementOf(SpecKey, tkgServiceVsphereKey, topologyKey, nodePoolKey)).([]interface{})[0].(map[string]interface{}); ok {
```
ramya-bangera, line 120: This looks very specific to TKGs and not supporting TKm vsphere?

internal/resources/cluster/resource_cluster.go
```diff
@@ -594,6 +668,26 @@ func resourceClusterInPlaceUpdate(ctx context.Context, d *schema.ResourceData, m
 
 		log.Printf("[INFO] cluster update successful")
 	}
+	// check default nodepool configuration update
+	npFullName := constructNpFullName(d)
```
ramya-bangera, line 672: While adding the support for update , do we need to add support for delete as well ? I am not sure if deleting the default nodepool is supported when there are additional node pools available. So it depends on if API supports the deletion of default nodepool

internal/resources/cluster/resource_cluster.go
```diff
@@ -563,6 +588,55 @@ func withMetaUpdate(d *schema.ResourceData, cluster *clustermodel.VmwareTanzuMan
...
+		nodepools := d.Get(helper.GetFirstElementOf(SpecKey, tkgVsphereClusterKey, topologyKey, nodePoolKey)).([]interface{})[0].(map[string]interface{})
+		poolSpec := nodepools[SpecKey].([]interface{})[0].(map[string]interface{})
+
+		if d.HasChange(helper.GetFirstElementOf(SpecKey, tkgVsphereClusterKey, topologyKey, nodePoolKey)) {
+			incomingWorkerNodeCount := poolSpec[workerNodeCountKey].(string)
+
+			if incomingWorkerNodeCount != "" {
+				nodepool.Spec.WorkerNodeCount = incomingWorkerNodeCount
+			}
+
```
ramya-bangera, line 632: A similar log message like L618 will be good to have

internal/resources/cluster/resource_cluster.go
```diff
@@ -563,6 +588,55 @@ func withMetaUpdate(d *schema.ResourceData, cluster *clustermodel.VmwareTanzuMan
...
+			return true
+		}
+	case nodepool.Spec.TkgVsphere != nil:
+		nodepools := d.Get(helper.GetFirstElementOf(SpecKey, tkgVsphereClusterKey, topologyKey, nodePoolKey)).([]interface{})[0].(map[string]interface{})
+		poolSpec := nodepools[SpecKey].([]interface{})[0].(map[string]interface{})
+
+		if d.HasChange(helper.GetFirstElementOf(SpecKey, tkgVsphereClusterKey, topologyKey, nodePoolKey)) {
+			incomingWorkerNodeCount := poolSpec[workerNodeCountKey].(string)
+
+			if incomingWorkerNodeCount != "" {
```
ramya-bangera, line 629: Is there a reason to limit the update just to worker node count here?

warroyo: This is what the nodepool resource does. so rather than adding functionality I used what was already existing. If we want to add functionality I can do that. https://github.com/vmware/terraform-provider-tanzu-mission-control/blob/main/internal/resources/cluster/nodepools/resource_node_pool.go#L646-L660

internal/resources/cluster/resource_cluster.go
warroyo: good catch, i missed that reference.

internal/resources/cluster/resource_cluster.go
warroyo: it does look like you can delete the default nodepool after another node pool is added. I'll look into how this might be done

warroyo: I think this would be a much bigger change since there is a check to ensure certain fields exist within the spec and the nodepool is one of them. this would require changing a lot more. I think we should table this for another update if it's needed.

internal/resources/cluster/resource_cluster.go
ramya-bangera: Yeah sound good, lets keep it the same as nodepool resource.

internal/resources/cluster/resource_cluster.go
ramya-bangera: Ok @willisren - couple of questions:
1.  Does the API implementation takes care of throwing error when the there is only one nodpool in the cluster (say default nodepool created at the time of cluster create) and user tries to delete it?
2. If there are multiple node pools and the user tries to update the default(or any) nodepool with worker node count of 0, is this operation supported via nodepool Update API? - what exactly happens in this scenario?

Also @warroyo - I understand adding delete support might require more changes so, can you let us know how soon this is required from TMC terraform provider? Is there an urgent customer requirement on this?

ramya-bangera: cc: @vmw-vjn

willisren: 1. Yes，there will be an error "the last node pool cannot be deleted" thrown out if users try to delete the last nodepool.
2. No, there will be an error "worker count is required to be greater than 0" thrown out if users try to update the replicas to 0 for a nodepool unless the autoscaler is enabled on it.

ramya-bangera (approved):

internal/resources/cluster/resource_cluster.go
```diff
@@ -563,6 +593,57 @@ func withMetaUpdate(d *schema.ResourceData, cluster *clustermodel.VmwareTanzuMan
 	return true
 }
 
+func withTKGNodePoolUpdate(d *schema.ResourceData, nodepool *nodepoolmodel.VmwareTanzuManageV1alpha1ClusterNodepoolNodepool) bool {
+	switch {
+	case nodepool.Spec.TkgServiceVsphere != nil:
+		nodepools := d.Get(helper.GetFirstElementOf(SpecKey, tkgServiceVsphereKey, topologyKey, nodePoolKey)).([]interface{})[0].(map[string]interface{})
+		poolSpec := nodepools[SpecKey].([]interface{})[0].(map[string]interface{})
+		// poolInfo := nodepools["info"].([]interface{})[0].(map[string]interface{})
```
shobha2626, line 601: Please remove the commented code

internal/resources/cluster/resource_cluster.go
```diff
@@ -563,6 +593,57 @@ func withMetaUpdate(d *schema.ResourceData, cluster *clustermodel.VmwareTanzuMan
...
+		}
+	case nodepool.Spec.TkgVsphere != nil:
+		nodepools := d.Get(helper.GetFirstElementOf(SpecKey, tkgVsphereClusterKey, topologyKey, nodePoolKey)).([]interface{})[0].(map[string]interface{})
+		poolSpec := nodepools[SpecKey].([]interface{})[0].(map[string]interface{})
+
+		if d.HasChange(helper.GetFirstElementOf(SpecKey, tkgVsphereClusterKey, topologyKey, nodePoolKey)) {
+			incomingWorkerNodeCount := poolSpec[workerNodeCountKey].(string)
+
+			if incomingWorkerNodeCount != "" {
+				nodepool.Spec.WorkerNodeCount = incomingWorkerNodeCount
```
shobha2626, line 635: only worked node count is updatable in TKGvSphere nodepool?

warroyo: this is what the current implementation allows through the nodepools resource. I did not change this functionality for this PR. We could add more updateable fields but I think that should be a second PR so we can get this feature delivered for the customer ASAP. see current implementation here https://github.com/vmware/terraform-provider-tanzu-mission-control/blob/main/internal/resources/cluster/nodepools/resource_node_pool.go#L646-L660

ramya-bangera (approved):

shobha2626 (approved):

PR #342, opened 2023-11-20
Feat/Permission Template -- Feature Complete

GilTeraSky:
    This PR includes the complete implementation of a Permission Template.
3. **Additional information**
    It was decided to go with a data source because there is no real state for a permission template as it can be generated for a 
    non-existing credentials.
    
    The following issues were addressed as follows:
    a. When generating a permission template for EKS, the templateValue field might contains values which are not defined as 
        parameters in the CloudFormation template. The workaround which was applied is to decode the template, and remove the 
        undefined values. This values are being kept instead in another computed field "undefined_template_values".
    b. The CloudFormation template returned from the API defined only a "Message" field as output and which contains text and 
        the ARN. The workaround should be implemented in the Terraform code to split the message data and get the ARN only.


shobha2626:
@vmw-vjn having a guide section for permission template would be helpful?

shobha2626:
Please squash the commits accordingly

internal/resources/permissiontemplate/tests/datasource_tf_configs.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+const (
+	dataProtectionPermissionTemplateDataSourceName = "data_protection_permissions"
+	eksPermissionTemplateDataSourceName            = "eks_permissions"
+)
+
+var (
+	dataProtectionPermissionTemplateDataSourceFullName = fmt.Sprintf("data.%s.%s", permissiontemplateres.ResourceName, dataProtectionPermissionTemplateDataSourceName) //nolint:unused
+	eksPermissionTemplateDataSourceFullName            = fmt.Sprintf("data.%s.%s", permissiontemplateres.ResourceName, eksPermissionTemplateDataSourceName)            //nolint:unused
```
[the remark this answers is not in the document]
GilTeraSky: For some reason the linter recognizes these variables as unused but they are used in the test file.

GilTeraSky:
> @vmw-vjn having a guide section for permission template would be helpful?

I included the entire Terraform config that utilizes the permission template, from my understanding a guide is needed when you need to create assets manually.
In this case it is very simple, you have a data source with certain fields to be used in another resource, therefore it is very straight forwards to implement, especially when you can see the entire Terraform.

vrabbi (dismissed):

shobha2626 (approved):
LGTM

internal/client/permissiontemplate/permission_template_resource.go
```diff
@@ -0,0 +1,79 @@
+/*
+Copyright © 2023 VMware, Inc. All Rights Reserved.
```
ramya-bangera, line 2: nit: update copyright to 2024

ankitsny: done

ramya-bangera (approved):

shobha2626 (approved):

PR #354, opened 2023-12-18
Models, clients, schema with CRUD support for provisioner

shobha2626:
1. **What this PR does / why we need it**: Models, clients, schema with CRUD support for provisioner. Manually tested the changes

        Fixes #

3. **Additional information**


ramya-bangera:
@shobha2626 - Raise the MR against feature branch to keep main stable

internal/resources/provisioner/data_source_provisioner.go
```diff
@@ -0,0 +1,63 @@
...
+	if !ok {
+		return diag.Errorf("unable to read management cluster name")
+	}
+
+	fn := &provisioner.VmwareTanzuManageV1alpha1ManagementclusterProvisionerFullName{
+		Name:                  provisionerName,
+		ManagementClusterName: managementClusterName,
+	}
+
+	resp, err := config.TMCConnection.ProvisionerResourceService.ProvisionerResourceServiceGet(fn)
```
ramya-bangera, line 48: The purpose of adding the datasource is that user can see all the existing provisioners in a management cluster, this Get API would just return the info of the specified provisioner and not list all the provisioners. So this should be replaced with List call and not GET, agree?

internal/resources/provisioner/resource_provisioner.go
```diff
@@ -0,0 +1,172 @@
...
+		log.Println("[ERROR] error while retrieving Tanzu auth config")
+		return diag.Errorf("error while retrieving Tanzu auth config")
+	}
+
+	provisionerName, ok := d.Get(NameKey).(string)
+	if !ok {
+		return diag.Errorf("unable to read provisioner name")
+	}
+
+	managementClusterName, ok := d.Get(ManagementClusterNameKey).(string)
```
ramya-bangera, line 72: Replace it with converter

internal/resources/provisioner/resource_provisioner.go
```diff
@@ -0,0 +1,172 @@
...
+var provisionerSchema = map[string]*schema.Schema{
+	NameKey: {
+		Type:        schema.TypeString,
+		Description: "Name of the provisioner",
+		Required:    true,
+		ForceNew:    true,
+	},
+	ManagementClusterNameKey: {
+		Type:        schema.TypeString,
+		Description: "Name of the management cluster",
```
ramya-bangera, line 47: here I feel its better to mention that it is currently not supported for TKG. @vmw-vjn any thoughts?

ramya-bangera (changes requested):

internal/resources/provisioner/resource_provisioner.go
vmw-vjn: First, I think any mention of applicability of resource - should in the docs (i.e.) docs/resources folder as md file. And not in code-comments in go file. Second - if we need to mention in code comments here, I'd suggest stating that TKG create & use default provisioners. As stated by LCM team in this Slack conversation: https://vmware.slack.com/archives/C0299HW9GH5/p1702905733958259

internal/resources/provisioner/data_source_provisioner.go
```diff
@@ -0,0 +1,63 @@
+/*
+Copyright © 2021 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
shobha2626: Done

internal/resources/provisioner/data_source_provisioner.go
shobha2626: Agree, add the data source support to list all the provisioners in a management cluster as well as to read individual provisioner also

internal/resources/provisioner/resource_provisioner.go
shobha2626: The schema for the provisioner is straight forward and is not nested as other resources. Adding the converter is not required which is comparatively more lines of changes than the current implementation, which is very easy to read and understand

internal/resources/provisioner/resource_provisioner.go
shobha2626: Agree with @vmw-vjn, we need to capture all these details in the document and not in the implementation side

ramya-bangera: @vmw-vjn  let me add more details.

1. These are node go code comments, instead it is the schema description which does appear in the provider documentation as a description of each of the field in the schema which user would refer to while constructing the resource definition. You can see an example in the repo [here](https://registry.terraform.io/providers/vmware/tanzu-mission-control/latest/docs/resources/integration#integration_name)
2. I agree that first of all this limitation should be stated in the TMC doc which we can easily refer to in TMC terraform provider docs. But unfortunately it is missing from there so somewhere in terraform provider doc it should be added to ensure we avoid more questions around the same when user tries to create provisioner on TKG management clusters.

internal/resources/provisioner/resource_provisioner.go
ramya-bangera: I don't think its true, converter would not add more lines of code, and while we have the existing converter logic that can be utilised make it a practice to use it instead of relying on the legacy method. This will also ensure that we use and enhance the converter for future use cases

shobha2626: We have to read provisioner and management cluster name alone from schema to pass it to the request. This is done as part of the changes in the code, adding the converter involves adding the converter map for the model and construct functions for the requests and the response.

Converter is useful in case of nested schemas where we have to construct and flatten the schema. Using converter in such case, helps avoid multi lines of code written. Which is not the case with provisioner resource

internal/resources/provisioner/resource_provisioner.go
shobha2626: There is a bug in the `tfdocplugins` as the description of the schema is not included in the doc generation. Anyways, adding clear description in the schema. 
We definitely should have this captured in doc as well

internal/resources/provisioner/resource_provisioner.go
```diff
@@ -0,0 +1,207 @@
...
+var provisionerSchema = map[string]*schema.Schema{
+	NameKey: {
+		Type:        schema.TypeString,
+		Description: "Name of the provisioner",
+		Required:    true,
+		ForceNew:    true,
+	},
+	ManagementClusterNameKey: {
+		Type:        schema.TypeString,
+		Description: "Name of the management cluster. Edit operation such as create, update and delete is not supported for TKGm & TKGs management cluster provisioners.",
```
ramya-bangera, line 47: Use official names

internal/models/provisioner/provisioner_create.go
```diff
@@ -0,0 +1,37 @@
+// Code generated by go-swagger; DO NOT EDIT.
```
ramya-bangera, line 1: Add copyright wherever missing

internal/resources/provisioner/data_source_provisioner.go
```diff
@@ -0,0 +1,84 @@
...
+	clienterrors "github.com/vmware/terraform-provider-tanzu-mission-control/internal/client/errors"
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/helper"
+	provisioner "github.com/vmware/terraform-provider-tanzu-mission-control/internal/models/provisioner"
+	"github.com/vmware/terraform-provider-tanzu-mission-control/internal/resources/common"
+)
+
+func DataSourceProvisioner() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: func(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+			return dataSourceProvisionerListRead(helper.GetContextWithCaller(ctx, helper.DataRead), d, m)
```
ramya-bangera, line 24: nit: better to rename this method, doesn't sound so great  may be - dataSourceProvisionersRead?

internal/resources/provisioner/resource_provisioner.go
shobha2626: Making use of converter

internal/resources/provisioner/resource_provisioner.go
shobha2626: Done

internal/models/provisioner/provisioner_create.go
shobha2626: Done

internal/resources/provisioner/data_source_provisioner.go
shobha2626: Done

shobha2626:
Closed the PR accidentally, please review the changes here - https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/357

PR #372, opened 2024-01-19
Add wait time for create/update operation to complete.

ramya-bangera:
      Adding a wait time of  10sec after the create/update operation as the operation takes some time to complete and there is no status field to rely on to check the completion of create/update operation.
       Issue tracked in internal jira

3. **Additional information**

Screenshot 2024-01-19 at 12 10 31 PM

Test logs from automation:

### DIFF
--- a/internal/resources/provisioner/resource_provisioner.go
+++ b/internal/resources/provisioner/resource_provisioner.go
@@ -8,6 +8,7 @@ package provisioner
 import (
 	"context"
 	"log"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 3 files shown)


internal/resources/provisioner/resource_provisioner.go
```diff
@@ -80,6 +86,9 @@ func resourceProvisionerCreate(ctx context.Context, d *schema.ResourceData, m in
 
 	d.SetId(provisionerResponse.Provisioner.Meta.UID)
 
+	log.Printf("Wait for %d seconds after the create the operation before fetching the state", waitTime)
+	time.Sleep(waitTime)
```
shobha2626, line 90: Wonder why the wait time is required to provisioner resource alone?

ramya-bangera: Please check the comment I have added in the L28

shobha2626: It's the same case with resources like cluster-group, namespace and workspace? We don't have status field in these resources?

ramya-bangera: In those cases the API returns a response is guaranteeing that resource is created but that's not the same case here. That's why we are seeing a provider crash! now I don't know why the API implementation is this way for provisioner alone so your question is specific to API behaviour which only the service owner can answer.

shobha2626 (approved):

ankitsny (approved):

PR #378, opened 2024-02-06
Update the doc link and add the example link to the cutom policy guide

ramya-bangera:
   Initially this check was added as the custom template creation via TMC TF provider was not supported. Now since the provider supports creation of the custom template removing this check,
    to let user create the custom template and assign it to custom policy on the fly.

        Fixes #

3. **Additional information**

Screenshot 2024-02-12 at 4 40 01 PM

Screenshot 2024-02-12 at 4 39 38 PM

### DIFF
--- a/docs/guides/tanzu-mission-control_policy.md
+++ b/docs/guides/tanzu-mission-control_policy.md
@@ -211,7 +211,7 @@ resource "tanzu-mission-control_iam_policy" "namespace_scoped_iam_policy" {
 }
 ```
 
-## Custom Policy on a CLuster Group
+## Custom Policy on a Cluster Group
 
 ```terraform
 /*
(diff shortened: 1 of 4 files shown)


docs/guides/tanzu-mission-control_policy.md
```diff
@@ -278,5 +278,82 @@ resource "tanzu-mission-control_custom_policy" "cluster_group_scoped_tmc-block-r
 ## Custom Template and Custom Policy
```
vmw-vjn, line 278: Section  *Custom Template and Custom Policy* seems to only have example for creating custom policy template. But not usage of said custom template in a custom policy.

And perhaps, section *Custom Policy on a Cluster Group* should be under this section? As an example of applying custom policy via Terraform.

vmw-vjn (commented):
Custom policy template and custom policy examples in doc seem to be scattered.

vmw-vjn (approved):

shobha2626 (approved):

PR #379, opened 2024-02-13
Fix Provisioner List Data read

shobha2626:
1. **What this PR does / why we need it**: Fix Provisioner List Data read

Data source acceptance test

image

{
  "version": 4,
  "terraform_version": "1.6.4",
  "serial": 2,
  "lineage": "0b6f3483-1df4-bc67-fa8e-0263fed0b3f9",
  "outputs": {},
  "resources": [
    {
      "mode": "data",
      "type": "tanzu-mission-control_provisioner",
      "name": "read_provisioner",
      "provider": "provider[\"vmware/devtest/tanzu-mission-control\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "id": "prvn:01HP45HJY9NJTTRB69BH1PZ49S",
            "provisioners": [
              {
                "management_cluster": "eks",
                "meta": [],
                "name": "demo-test",
                "org_id": ""
              }
            ]
          },
          "sensitive_attributes": []
        }
      ]
    }
  ],
  "check_results": null
}

// LIST of Provisioners

"provisioners": [
              {
                "management_cluster": "eks",
                "meta": [],
                "name": "tf-test",
                "org_id": ""
              },
              {
                "management_cluster": "eks",
                "meta": [],
                "name": "tf-prv-test-2322938641799613315",
                "org_id": ""
              },
              {
                "management_cluster": "eks",
                "meta": [],
                "name": "tf-prv-test-9059343843537634380",
                "org_id": ""
              },
              {
                "management_cluster": "eks",
                "meta": [],
                "name": "tf-prv-test-2312083577691511843",
                "org_id": ""
              },

4. **Additional information**

5. **Special notes for your reviewer**


internal/resources/provisioner/data_source_provisioner.go
```diff
@@ -28,35 +31,48 @@ func DataSourceProvisioner() *schema.Resource {
...
+	var resp *provisioner.VmwareTanzuManageV1alpha1ManagementclusterProvisionerListprovisionersResponse
+
+	model, err := tfModelDataConverter.ConvertTFSchemaToAPIModel(d, []string{provisionerKey, nameKey, managementClusterNameKey})
+	if err != nil || model == nil {
 		return diag.FromErr(errors.Wrapf(err, "Couldn't read Tanzu Mission Control provisioner configurations."))
 	}
 
-	if model.FullName.Name == "" {
-		resp, err := config.TMCConnection.ProvisionerResourceService.ProvisionerResourceServiceList(model.FullName)
+	if model.Provisioners[0].FullName.Name == "" {
```
ramya-bangera, line 74: before accessing the element in array like `model.Provisioners[0]`, add a check to validate list is not empty.

internal/resources/provisioner/data_source_provisioner.go
```diff
@@ -83,12 +91,38 @@ func dataSourceProvisionerRead(ctx context.Context, d *schema.ResourceData, m in
...
+		resp = p
+	}
+
+	err = tfModelDataConverter.FillTFSchema(resp, d)
+
+	if err != nil {
+		return diag.FromErr(errors.Wrapf(err, "Unable to populate tf schema"))
 	}
 
+	for _, prov := range resp.Provisioners {
```
ramya-bangera, line 121: With this logic , If you invoke the datasource multiple times can it create multiple entries in the state file? because
1.  the response might not maintain the order of provisioners 
2. A new provisioner is added then id will definitely change with this new field. Can you test this part?

ramya-bangera (changes requested):
1. Update the datasource example to reflect this change and documentation update accordingly, which I don't see in this change.
2. Run the datasource acceptance test and attach the screenshot

internal/resources/provisioner/data_source_provisioner.go
shobha2626: Have added the check to see if the model is nil or not in line 70

internal/resources/provisioner/data_source_provisioner.go
shobha2626: On invoking datasource multiple times, if there is a change in the provisioners state from TMC server side (that is a new create/delete/update), the unique ID changes but it does not create multiple entries. I have tested the same

internal/resources/provisioner/data_source_provisioner.go
ramya-bangera: better to add  a nil check for provisioner too `model.Provisioners == nil`

ishangupta-ds (approved):

ramya-bangera (approved):

PR #390, opened 2024-03-22
Refresh auth context while polling for data protection resources

rnarenpujari:
1. **What this PR does / why we need it**:

The data protection related resources have a poll stage where post-creation they wait to reach a terminating state. In Self Managed environments, the auth token may expire during this process. This PR adds refresh logic.

Manually tested the fix.

Acceptance tests passed:

Screenshot 2024-03-22 at 8 26 52 PM

2. **Which issue(s) this PR fixes**

        (optional, in `fixes #<issue number>` format, will close the issue(s) when PR gets merged):

        Fixes #

3. **Additional information**

4. **Special notes for your reviewer**

PR #393, opened 2024-03-26
Refresh auth context while polling for clustergroup backupschedule

rnarenpujari:

Missed one case in this earlier PR: https://github.com/vmware/terraform-provider-tanzu-mission-control/pull/390

The acceptance tests for this resource are failing for a different reason and that is being tracked separately. This is a minor change and should not cause any regression.

        Fixes #

3. **Additional information**


ankitsny (approved):

internal/resources/backupschedule/resource_backup_schedule.go
```diff
@@ -507,6 +507,11 @@ func readCGResourceWait(ctx context.Context, config *authctx.TanzuContext,
 			resourceFullName)
 
 		if err != nil || resp == nil || resp.Schedule == nil {
+			if clienterrors.IsUnauthorizedError(err) {
+				authctx.RefreshUserAuthContext(config, clienterrors.IsUnauthorizedError, err)
+				continue
+			}
+
 			return nil, err
```
ankitsny, line 515: @rnarenpujari, can we add a unit test for this?

rnarenpujari: Currently we only have acceptance tests and simulating this condition seems difficult. Looks like other resources also have mock tests which would be ideal here - but it doesn't seem the contractors added those when writing the data protection resources. Since it might be a sizeable effort, would prefer to track this separately. But let me know if you feel strongly about adding them with this PR itself.

eastaga (approved):
lgtm

internal/resources/backupschedule/resource_backup_schedule.go
rnarenpujari: Will track the addition of mock tests for this resource separately.

